PR #568, opened 2015-01-21
Iterator: skip keys which extended header can't be extracted

shaitan:
In our reality some records can be corrupted because of many issues. Iteration should skip such records if it is possible and continue iterating all valid records.  In this case it will safely skip records with corrupted extended header. While recovering all such keys will be recovered if at least one replica has a non-corrupted copy of these keys.

### DIFF
--- a/example/eblob_backend.c
+++ b/example/eblob_backend.c
@@ -106,6 +106,7 @@ struct eblob_backend_config {
 static int blob_iterate_callback_common(struct eblob_disk_control *dc, void *data, void *priv, int no_meta) {
 	struct dnet_iterator_ctl *ictl = priv;
 	struct dnet_ext_list elist;
+	struct eblob_backend_config *c = ictl->iterate_private;
 	uint64_t size;
 	int err;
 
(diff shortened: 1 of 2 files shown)


bioothod:
Does this mean that records with corrupted header will not be recovered?
But they can be read, do they? If they can be read, it might be better not to skip, but to set default ext header (zero timestamps and flags)

shaitan:
If some other replica has these key non-corrupted, recovery rewrites corrupted replicas with valid one. At the moment Elliptics will fail reading keys with header corruption. Is it correct to ignore header corruption while reading the key? Is header corruption signaling that whole record is corrupted, isn't it?

shaitan:
Also I think we can make recovery to cleanup keys which have only corrupted records in replicas.

bioothod:
If another record in different group has correct header, it will have higher timestamp and data will be successfully recovered even if we set default header for corrupted record.

Removing keys with invalid header isn't bad idea, but I'm afraid that there might be a bug introduced somewhere between non-ext-header and ext-header versions, which ends up with all non-ext-header keys to be treated as keys with invalid header.

How many keys with broken headers compared to keys without header do you have?

shaitan:
Updated changes:
 * iterator with `no-meta` returns zero metadata: zero timestamp and zero flags
 * iterator does not skip records with corrupted header but uses empty header for such keys

Issue #570, opened 2015-01-23
recovery broken? 


agend:
Gents,
  Tried last version and can't perform simple merge recovery. dnet_recovery script just stuck and nothing else. Did you test it?

bioothod:
Please me more specific, in particular what does debug log level tell, what does stat file contain?
@shaitan please take a look

agend:
@bioothod @shaitan Please take a look: https://yadi.sk/d/QEYV_msgeC79w
Also I can't stop it via CTRL-C

agend:
Guys, Any new information on this?

shaitan:
Hi @agend. I'm investigating your issue and will make a fix today.

agend:
@shaitan Thanks.

shaitan:
@agend which versions of eblob and elliptics did you use at the test?

agend:
@shaitan eblob: 0.22.10 elliptics: 2.26.3.35

agend:
Still no luck. Now it's can't find any keys to recover. Logs: https://yadi.sk/d/N0dNADd2eH2yu

shaitan:
Is there any key that should be recovered? Can you describe what you do before recovering?

agend:
1) Started empty node1
2) Filled with 20k records
3) Started one more empty node in the same group
4) Run recovery

shaitan:
Could you provide logs of 2) stage?

agend:
@shaitan Here you go: https://yadi.sk/d/f_qChYvoeJS7N

shaitan:
@agend there is no write command at the log. Can you please rerun your case and provide client log while it writes 20k records and log of recovering?

agend:
@shaitan  I have just checked file https://yadi.sk/d/f_qChYvoeJS7N and found this:
2015-01-28 19:45:12.735054 0000000000000000/3131/3087 NOTICE: 3d53caf3669af025: EBLOB: blob-write: WRITE: Ok: offset: 0, size: 10. 'backend_id': 1
2015-01-28 19:45:12.735085 0000000000000000/3107/3087 INFO: 0:0000000000000000...0000000000000000: ROUTE_LIST: client: 127.0.0.1:46501, trans: 1, cflags: 0x18 [direct|nolock], time: 379 usecs, err: 0. 

I think this one is write command or I missing something?

agend:
@bioothod Or this 
[...]

PR #571, opened 2015-01-27
recovery: fixed stucking `dnet_recovery merge` on keys which are not needed to be recovered

shaitan:


bioothod:
Should I make a new release, or there are more pull requests pending?

shaitan:
You can because it fixed https://github.com/reverbrain/elliptics/issues/570. Please, bump the version of Eblob in Elliptics dependencies because in Elliptics I removed filtering keys on Elliptics side because it was moved to Eblob side.

PR #572, opened 2015-02-11
statistics: restored `backend::config` for disabled eblob and filesystem backends.

shaitan:
Provided `to_json` methods for eblob and filesystem config templates which serializes to json parsed config values. Made `fill_disabled_backend_config` to use `to_json` method of config template for filling `backend::config` with parsed config values if `to_json` is available, otherwise to fill `backend::config_template` with unparsed values from original config.

### DIFF
--- a/monitor/backends_stat_provider.cpp
+++ b/monitor/backends_stat_provider.cpp
@@ -38,7 +38,8 @@ backends_stat_provider::backends_stat_provider(struct dnet_node *node)
  */
 static void fill_backend_backend(rapidjson::Value &stat_value,
                                  rapidjson::Document::AllocatorType &allocator,
-                                 const struct dnet_backend_io &backend) {
+                                 const struct dnet_backend_io &backend,
+                                 const dnet_backend_info &config_backend) {
 	char *json_stat = NULL;
 	size_t size = 0;
 	struct dnet_backend_callbacks *cb = backend.cb;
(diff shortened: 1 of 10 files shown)


monitor/backends_stat_provider.cpp
```diff
@@ -188,6 +216,10 @@ static rapidjson::Value& backend_stats_json(uint64_t categories,
 		fill_disabled_backend_config(stat_value, allocator, config_backend);
 	}
 
+	if (categories & DNET_MONITOR_BACKEND) {
```
bioothod: This has to be moved into fill_disabled_backend_config() function, since if there is no json generated, there will be no ["backend"]["config"] section, and it will crash

shaitan: Thank you, I really forgot about it. I'm making a commit with fix.

shaitan: Fixed: 70680d5e5012fd65fe778bd817c0f850485ce877

bioothod:
Are there any other changes pending (in particualal, related to https://github.com/reverbrain/elliptics/issues/570 issue) , or should I make a new version?

shaitan:
https://github.com/reverbrain/elliptics/issues/570 is still under investigation. And for the new version I'm preparing another one pull request.

PR #573, opened 2015-02-20
Top

abudnik:
There exists a problem of the periodic network overload to elliptics node. Network load is created by frequent access to some subset of keys or by accessing very big keys. If it was possible to get list of keys, which have most impact on network traffic size, then these keys could be moved to other caching machines, decreasing total traffic size to nodes.

To solve task of searching such keys it is necessary to rank all keys, which were accessed, by their impact on generated traffic. Moreover, it is necessary to use only keys that where accessed during last N seconds. To solve task of keys caching we do not need all keys in list, but only limited part of keys, which have most impact on traffic. Let's call such list - top keys statistic.

This patch adds ability to get top keys statistic. It is also possible to extract top keys statistic by parsing of logs, but it will cause additional system load. Load testing have shown that the version of dnet_ioserv with top keys statistic did not lead to any significant divergence in the number of RPS or latency values with the version without top, where the test was a small data querieng (<= 1 Kb) at 80k RPS to one node.

File example/monitoring-stats.qdoc describes configuration of the top keys statistics module.


include/elliptics/interface.h
```diff
@@ -338,11 +338,6 @@ struct dnet_config
 	/* Config file name for handystats library */
 	const char 	*handystats_config;
 
-	/*
-	 * Monitor socket port
-	 */
-	unsigned int		monitor_port;
-
```
bioothod: This breaks API and ABI

monitor/event_stats.hpp
```diff
@@ -0,0 +1,244 @@
...
+#include <algorithm>
+#include <functional> // not2
+#include <vector>
+#include <stack>
+
+namespace ioremap { namespace monitor {
+
+class mutex_lock_policy
+{
+	typedef std::mutex mutex_t;
```
bioothod: Please do not introduce such obscure types.
If you saw them in the code already, it would be better to change that

monitor/event_stats.hpp
```diff
@@ -0,0 +1,244 @@
...
+public:
+	class unique_lock {
+	public:
+		unique_lock(mutex_lock_policy *policy)
+		: lock_(policy->get_lock())
+		{}
+	private:
+		std::unique_lock<mutex_t> lock_;
```
bioothod: This doesn't follow c++ conventions. member fields are called m_lock

monitor/event_stats.hpp
```diff
@@ -0,0 +1,244 @@
...
+};
+
+class null_lock_policy
+{
+public:
+	class unique_lock {
+	public:
+		unique_lock(null_lock_policy *policy) { (void) policy; }
+	};
+};
```
bioothod: Why do you need this hand-coded unique lock?

monitor/event_stats.hpp
```diff
@@ -0,0 +1,244 @@
...
+ * significant amount of events multiplied by event size exceeded memory limitations.
+ * Therefore statistics are computed approximately in case of lots of heavyweight events.
+ *	Any type E of event must support following methods: (get/set)_weight, (get/set)_frequency,
+ * and (get/set)_time.
+ *	Currently only get_top method implemented, which could be used for getting top k events
+ * with highest weight, where weight is approximate sum of weight of particular events during
+ * observable period of time (period_in_seconds).
+ */
+template<typename E, typename lock_policy = null_lock_policy>
+class event_stats : private lock_policy
```
bioothod: No.

monitor/event_stats.hpp
```diff
@@ -0,0 +1,244 @@
...
+		typename lock_policy::unique_lock lock(this);
+		auto it = treap.find(event.get_key());
+		if (it) {
+			update_weight(*it, time, period, event.get_weight());
+			update_frequency(*it, time, period, 1.);
+			it->set_time(time);
+			treap.decrease_key(it);
+		} else {
+			if (num_events < max_events) {
+				treap.insert(new E(event));
```
bioothod: What will happen, if this fails?

monitor/event_stats.hpp
```diff
@@ -0,0 +1,244 @@
...
+	 * \a top_size - result container, that should support push_back operation and random access iterators
+	 */
+	template< typename ResultContainer >
+	void get_top(size_t k, time_t time, ResultContainer &top_size)
+	{
+		std::vector< typename treap_t::p_node_type > top_nodes;
+
+		{
+			typename lock_policy::unique_lock lock(this);
+			treap_to_container(treap.top(), top_nodes);
```
bioothod: How elements are sorted, and therefore what is complexity of this operation?
Does it imply weight-unsorted array, which is being sorted every time we read it?

monitor/monitor.cpp
```diff
@@ -27,9 +27,19 @@
...
 #include "backends_stat_provider.hpp"
 #include "procfs_provider.hpp"
+#include "top_provider.hpp"
 
-static ioremap::monitor::monitor* get_monitor(struct dnet_node *n) {
-	return n->monitor ? static_cast<ioremap::monitor::monitor*>(n->monitor) : NULL;
+#include "../example/config.hpp"
+
+static ioremap::monitor::monitor_config* get_monitor_config(struct dnet_node *n) {
+	using namespace ioremap::elliptics::config;
```
bioothod: Do not use 'using', it doesn't save a dime here

monitor/monitor.hpp
```diff
@@ -24,11 +24,28 @@
...
 namespace ioremap { namespace monitor {
 
+struct monitor_config
+{
+	unsigned int	monitor_port;
+	bool			has_top;
+	size_t			top_length;
+	size_t			events_size;
+	int				period_in_seconds;
+
```
bioothod: Broken indent

monitor/statistics.cpp
```diff
@@ -190,27 +190,14 @@ statistics::statistics(monitor& mon, struct dnet_config *cfg) : m_monitor(mon)
 
 void statistics::add_provider(stat_provider *stat, const std::string &name)
 {
-	std::unique_lock<std::mutex> guard(m_provider_mutex);
-	m_stat_providers.emplace_back(std::unique_ptr<stat_provider>(stat), name);
+	boost::unique_lock<rw_lock> guard(m_provider_lock);
```
bioothod: No boost is allowed

monitor/statistics.hpp
```diff
@@ -28,6 +28,10 @@
 #include <mutex>
 #include <sstream>
 #include <thread>
+#include <map>
+
+#include <boost/thread/locks.hpp>
+#include <boost/thread/shared_mutex.hpp>
```
bioothod: Boost is not allowed.

monitor/top_provider.cpp
```diff
@@ -0,0 +1,85 @@
...
+namespace ioremap { namespace monitor {
+
+top_provider::top_provider(struct dnet_node *node)
+: m_node(node)
+{
+	auto monitor = get_monitor(node);
+	m_top_stats = monitor->get_top_stats();
+}
+
+static inline char *dnet_dump_id_str_full(const unsigned char *id)
```
bioothod: Move this function to other dump functions family

monitor/event_stats.hpp
```diff
@@ -0,0 +1,244 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Elliptics.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef EVENT_STATS_HPP
+#define EVENT_STATS_HPP
```
bioothod: Please use common guard names __DNET_MONITOR_EVENT_STATS_HPP

monitor/top_stats.hpp
```diff
@@ -0,0 +1,125 @@
...
+}}  /* namespace ioremap::monitor */
+
+namespace ioremap { namespace cache {
+
+template<>
+struct treap_node_traits<ioremap::monitor::key_stat_event>
+{
+	typedef const struct dnet_id* key_type;
+	typedef size_t priority_type;
+};
```
bioothod: This is being repeated, can it be merged into single trait structure?

monitor/top_stats.hpp
```diff
@@ -0,0 +1,125 @@
...
+	typedef ioremap::cache::treap_node_traits<key_stat_event>::priority_type priority_type;
+
+	key_type get_key() const { return &m_id; }
+	priority_type get_priority() const { return m_last_access; }
+
+	inline static int key_compare(const key_type &lhs, const key_type &rhs) {
+		return dnet_id_cmp(lhs, rhs);
+	}
+
+	inline static int priority_compare(const priority_type &lhs, const priority_type &rhs) {
```
bioothod: This set of methods is being repeated too

tests/pytests/test_monitor_top.py
```diff
@@ -0,0 +1,81 @@
...
+
+def check_keys_fields(keys):
+    for k in keys:
+        assert k['group']
+        assert k['id']
+        assert k['size']
+        assert k['frequency']
+
+class TestMonitorTop:
+    def test_single_write(self, server, simple_node):
```
bioothod: No, this is not a top test at all.
Please write a real test where many packets are sent, statistics is being calculated and checked

tests/run_servers.cpp
```diff
@@ -164,15 +164,33 @@ static void setup_signals()
 	sigprocmask(SIG_UNBLOCK, &sa.sa_mask, NULL);
 }
 
-static bool read_option(const rapidjson::Value &doc, const std::string &value, bool default_value)
+template<typename T>
+void read_option(const rapidjson::Value &doc, const std::string &value, T default_value, T& result);
+
+template<>
+void read_option(const rapidjson::Value &doc, const std::string &value, bool default_value, bool& result)
```
bioothod: Why is this needed?

tests/stats_test.cpp
```diff
@@ -0,0 +1,390 @@
...
+
+}  /* namespace tests */
+
+namespace ioremap { namespace cache {
+
+template<>
+struct treap_node_traits<tests::test_event>
+{
+	typedef const std::string* key_type;
+	typedef size_t priority_type;
```
bioothod: This is being repeated again. Please refactor.

tests/stats_test.cpp
```diff
@@ -0,0 +1,390 @@
...
+namespace tests {
+
+static std::shared_ptr<nodes_data> global_data;
+
+static void destroy_global_data()
+{
+	global_data.reset();
+}
+
+class test_event : public ioremap::cache::treap_node_t<test_event> {
```
bioothod: This looks like copy-pasted class

tests/stats_test.cpp
```diff
@@ -0,0 +1,390 @@
...
+		server_config::default_value().apply_options(config_data()
+		    ("group", 5)
+			("monitor_top", top_params)
+		)
+	}), path);
+
+	global_data = start_nodes(start_config);
+}
+
+static void test_top_provider_existance()
```
bioothod: existEnce

tests/stats_test.cpp
```diff
@@ -0,0 +1,390 @@
...
+template<typename container_t>
+static size_t events_symmetric_diff(container_t &fst, container_t &snd)
+{
+	std::sort(fst.begin(), fst.end(), test_event::key_compare_event);
+	std::sort(snd.begin(), snd.end(), test_event::key_compare_event);
+
+	std::vector<test_event> diff(fst.size() + snd.size());
+	auto it = std::set_symmetric_difference(fst.begin(), fst.end(),
+											snd.begin(), snd.end(),
+											diff.begin(), test_event::key_compare_event);
```
bioothod: Indent problem

tests/stats_test.cpp
```diff
@@ -0,0 +1,390 @@
...
+	BOOST_REQUIRE_EQUAL(events_symmetric_diff(top_events, result), 0);
+}
+
+bool register_tests(test_suite *suite, node n)
+{
+	(void) n;
+	ELLIPTICS_TEST_CASE_NOARGS(test_top_provider_existance);
+	ELLIPTICS_TEST_CASE_NOARGS(test_event_stats_boundary_conditions);
+	ELLIPTICS_TEST_CASE_NOARGS(test_event_stats_no_time_dependency);
+	ELLIPTICS_TEST_CASE_NOARGS(test_event_stats_with_time_dependency);
```
bioothod: Am I right, that above tests show the difference between 2 copy-pasted implementations of the same algorithm?
How the heck can they differ?

tests/test_base.cpp
```diff
@@ -234,6 +252,19 @@ struct json_value_visitor : public boost::static_visitor<>
...
 	rapidjson::MemoryPoolAllocator<> *allocator;
 
+	void operator() (const config_data &value) const
+	{
+		rapidjson::Value result;
+		result.SetObject();
+
+		for (auto it = value.cbegin(); it != value.cend(); ++it) {
+			json_value_visitor visitor(it->first.c_str(), &result, allocator);
+			boost::apply_visitor(visitor, it->second);
```
bioothod: We only use boost-tests, nothing more from boost is allowed

tests/test_base.hpp
```diff
@@ -130,6 +130,10 @@ class server_config;
 
 class config_data
 {
+protected:
+	typedef boost::variant<std::vector<std::string>, std::string, bool, int64_t, config_data> variant;
```
bioothod: No boost is allowed except from boost-test

bioothod:
Overall idea is interesting, but pull request contains fair amount of things unrelated to original proposal. Please split statistics update part at least. Also various cleanups should not live here, but in a separate PR.

Tests are unacceptable, either explain them or rewrite.

Algorithm requires more description and examples from real life scenarios.

Things like private inheritance are way too ugly for simple updater class. Also, consider atomic variables if they are appropriate for counters.

monitor/statistics.hpp
abudnik: > grep -R "boost\/" ./elliptics | wc -l
>71
>grep -R "boost\/" ./elliptics/monitor/ | wc -l
>9

bioothod: The only real places where boost is needed are tests and python.
My initial implementation used boost intrusive lists, but I thought it was changed.

Anyway, do not add new dependencies, it would be great to get rid of current use cases, they are quite trivial like boost::lexical_cast (we have much faster helper) and the like

monitor/top_stats.hpp
abudnik: ioremap::cache::treap_node_traits used in ioremap::cache::treap_node_t
This is concrete template class specialization for type ioremap::monitor::key_stat_event:
> typedef const struct dnet_id* key_type; // for ioremap::monitor::key_stat_event
> typedef const uint8_t* key_type; // for ioremap::cache::data_t

Different types, so it is not possible to merge it into single data structure.

tests/pytests/test_monitor_top.py
abudnik: This test checks simple protocol agreements. All top statistics internal logic tested in cpp tests:
tests/stats_test.cpp

tests/stats_test.cpp
abudnik: test_event_stats_no_time_dependency - all events occurs in one moment
test_event_stats_with_time_dependency - some events may have different time of occurance

And test logic of both methods is not overlapped

tests/stats_test.cpp
```diff
@@ -0,0 +1,390 @@
...
+
+	return suite;
+}
+
+}
+
+int main(int argc, char *argv[])
+{
+	atexit(tests::destroy_global_data);
+	srand(time(0));
```
bioothod: Is this a good idea?

monitor/event_stats.hpp
abudnik: `template<typename E, typename lock_policy = null_lock_policy>
class event_stats : private lock_policy {...};`

event_stats used in the same way as stl containers used. But stl containers doesn't provide native thread-safety.
It is convinient to use thread-safety policy as an event_stats class template parameter.
Tests use null_lock_policy, and top_provider use mutex_lock_policy

tests/stats_test.cpp
bioothod: What is 'test_event' class for? Why does it copy-paste content of the cache processing?
What do tests do in details? Please describe it with more information about design of the tests, results, what is *not* supposed to be found in tests and what has to be there and so on 

From the first glance it looks like it compares several small (short-living) event sets fed into two event processors which are the same.

Do you have test data where you fed gigabytes of data, thousands of keys into the system and check for real that returned statistics matches the most actively downloaded keys?

monitor/event_stats.hpp
abudnik: It is difficult to say where it can fail definitely.

monitor/event_stats.hpp
bioothod: stl containers are simple and nice because they do not have private inheritance from the automatic locking machine, and it was done for purpose.

Do not do this. It would much simpler just to put fine-grained locks into statistics classes, or use atomics where appropriate.

monitor/event_stats.hpp
bioothod: Will it leak or crash?

monitor/event_stats.hpp
abudnik: Ok, I will move locking logic into event_stats impl. and remove locking policies.
Critical section should be as short as possible in add_event/get_top, because arised lock-contention degrades performance dramatically at high rps rates.

monitor/event_stats.hpp
abudnik: First of all I must mention, that event_stats designed as data structure with frequent add_event and rare get_top operations access pattern. So add_event must be as fast as possible to not affect overall performance.

Complexity of this operation (get_top) depends on number of expired events (events with timestamp value outside window size) - E, and total number of events - N.
E <= N

It does some operations under lock:
> O(N) - treap_to_container()
> O(E * logN) - worst case for sequental treap.erase()

And after unlock:
> std::partial_sort() // standart guarantees O(N * log k)
It is also possible to make heap of k elements and do (N-k) heap.update_min() operations, resulting in O(k + (N-k)*log k) = O((N-k)*log k), but with a small k complexity of this method is not significantly faster than std::partial_sort().

abudnik: >Does it imply weight-unsorted array, which is being sorted every time we read it?
Yes. get_top result depends on time_t time parameter, because array contents may change every time get_top is called.

abudnik: std::partial_sort() can be also replaced by next algorithm: build heap O(N) and then do O(k * logN) pop_heap(). It may be faster than std::partial_sort, but this method requires more code.
event_stats.get_top() called rarely, so optimization will case in very small performance boost.

monitor/event_stats.hpp
abudnik: No

monitor/top_stats.hpp
abudnik: This methods are part of treap_node_t interface.

monitor/event_stats.hpp
bioothod: Ok, I see. Please describe performance bits in more details in comments and documentation

monitor/statistics.hpp
abudnik: shared_mutex - is a wrapper around pthread_rwlock_t, so statistics::report could be accessed without exclusive locks. Also std::shared_mutex has appeared in C++14.

Ok, I will remove this. Will be slower, but without boost.

tests/stats_test.cpp
abudnik: >Why does it copy-paste content of the cache processing?

It has nothing to share with cache processing, but common treap_node_t interface.

>Do you have test data where you fed gigabytes of data, thousands of keys into the system and check for real that returned statistics matches the most actively downloaded keys?

Yes, it has been tested in the following way:
1) Loaded keys with random distribution of key sizes ranging from 100s of bytes to 10s of megabytes
2) Randomly accessed those keys, while waching results: http://sata17h.xxx.yandex.net:20000/top
dnet_ioserv is able to return a list of top keys with approximate traffic size and access frequency. So really heavy keys and keys with frequent access where found here in that list, what is expected.

monitor/statistics.hpp
bioothod: Thank you.

As per locks - shared locks is the first sign that locking is not fine-grained, and instead of data some piece of code is guarded. It is **always** a red flag to see whether there are design problems.

Also, we can not use c++14 since Ubuntu Precise and older RHELs which we support do not have this compiler.
You said that statistics reads are rare, thus it should not be a problem.

tests/stats_test.cpp
bioothod: Please highlight this in documentation, preferably with examples, uploaded key sizes, time of reads, stats output and so on - kind of tutorial for those who never saw the code on how to use this new feature

abudnik: >describe it with more information about design of the tests

event_stats - is a core abstract data type used in top keys statistics logic. If event_stats logic is correct, then top keys statistics will be correct too.
event_stats ADT provides statistics logic for ANY type, that supports methods {set/get}_weight, {set/get}_time, {set/get}_frequency.

test_top_provider_existence - if "top" section in "monitor" section in config exists and well-defined, then top_provider object must exists.

test_event_stats_boundary_conditions - various boundary conditions test, e.g. inserting more events than event_stats::m_max_events (== EVENTS_SIZE) will fit into event_stats max capacity (get_top() should never return more than m_max_events events).

test_event_stats_no_time_dependency - test that checks (binary) heap properties of event_stats ADT, e.g. monotonically increment event's weight within events_limit and then see if heaviest event occurs in get_top() result. All inserted events have same time of occurance.

test_event_stats_with_time_dependency - checks that event's weight and event's existence in event_stats container depends on time as it was designed (satisfy heap property and "old" events are not returned by get_top()).

I will add more comments in test_stats.cpp in next (fixing) PR.

PR #575, opened 2015-03-04
mmap elimination

abudnik:
This PR connected with appropriate eblob PR: https://github.com/reverbrain/eblob/pull/104


example/eblob_backend.c
```diff
@@ -86,15 +86,28 @@ static int eblob_read_params_compare(const void *p1, const void *p2)
...
 	struct eblob_backend_config *c = ictl->iterate_private;
 	uint64_t size;
+	void *data, *saved_data;
 	int err;
 
 	assert(dc != NULL);
-	assert(data != NULL);
+
+	size = dc->disk_size;
+	saved_data = data = malloc(size);
```
bioothod: This is not going to work with large files.
Why wouldn't you use new fd checksumming API?

library/crypto/sha512.c
```diff
@@ -321,6 +324,72 @@ sha384_stream (FILE *stream, void *resblock)
   return 0;
 }
 
+/* Compute SHA512 message digest for bytes read from file descriptor.
+   The resulting message digest number will be written into the 64
+   bytes beginning at RESBLOCK.  */
+int
+sha512_file_ctx (int fd, off_t offset, size_t count, struct sha512_ctx *ctx)
```
bioothod: This function is copy-pasted from eblob code, can it be called from library?

example/eblob_backend.c
abudnik: From library/dnet.c: dnet_iterator_callback_common:
position = combined = malloc(size);

But now I see that it allocates memory only when DNET_IFLAGS_DATA.
So it is possible to pass fd to this function and avoid extra memcpy.

In current version it (dnet_iterator_callback_common) may allocate memory enough to handle whole file data.

bioothod: It could be great to eliminate allocation

library/crypto/sha512.c
abudnik: sha512_file_ctx (elliptics) and sha512_file (eblob) - have different interfaces and implementation.
Another question, why this crypto library copy pasted in both elliptics and eblob? It seems that it has copy, because elliptics originally supported multiple backend implementations.

example/eblob_backend.c
abudnik: I found this in current code example/eblob_backend.c at blob_iterate_callback_common:

	if (dc->flags & BLOB_DISK_CTL_EXTHDR) {
		size -= sizeof(struct dnet_ext_list_hdr);
	}

It seems that size decremented, but data pointer is not adjusted. Is it ok or should I add
data += sizeof(struct dnet_ext_list_hdr);
here?

library/crypto/sha512.c
bioothod: Elliptics still supports multiple backends, but filesystem backend depends on eblob now - it stores key metadata there (like io user flags).

We can assume that eblob is always present in the system - elliptics depends on it forever so far, thus we could use the same sha implementation.

abudnik: I will move crypto lib from elliptics to eblob in a separate PR.

example/eblob_backend.c
abudnik: Done. See https://github.com/abudnik/elliptics/commit/70df7a83e585567e2862434596e76b83116d9ac8

library/crypto/sha512.c
abudnik: How about git submodule for crypto lib in both elliptics and eblob?

bioothod: No submodules - it brings horrible packaging and versioning problems, we already tried that and eventually moved every submodule into own package.
I used openssl first, but it has multithreading problems, so I put sha512 implementation into eblob from some FSF package, since it was only present in ubuntu and not in older rhel

abudnik: Install additional header (with different header, copyrights,etc) in eblob?

```cmake
install(FILES
    include/eblob/eblob.hpp
    include/eblob/blob.h
    include/eblob/sha512.h
    DESTINATION include/eblob/
    )
```

bioothod: Yes, that's ok. It also needs u64.h header

abudnik: This task (moving crypto lib to eblob) indirectly connected with mmap elimination.
I will do this in a separate PR (as a separate task), as I mentioned above.

bioothod: Okay

abudnik: But elliptics client will be dependent on eblob.
session::transform() uses dnet_local_digest_transform() from crypto.c where called sha512_process_bytes()

bioothod: Argh, crap, that's a show stopper I think.
Can we at least have exactly the same crypto dirs in both elliptics and eblob sources with detailed description of why they can not be merged?

shaitan: To ensure identity of crypto in both elliptics and eblob we should move crypto in separate package.

abudnik: I've decided merge both crypto libs.
Can I remove sha384 functions? They are not used.

bioothod: Sure, drop all unused bits

abudnik: https://github.com/abudnik/elliptics/commit/48ed6ada0434732dcb1d3323aee5debc75c8588e
https://github.com/abudnik/eblob/commit/864c3f77526aa8f154e0cb18b5283de5efdefc66

>$> cmp eblob/library/crypto/sha512.c elliptics/library/crypto/sha512.c && echo $?
>$> 0
>$> cmp eblob/library/crypto/sha512.h elliptics/library/crypto/sha512.h && echo $?
>$> 0

PR #576, opened 2015-03-19
Fixes & optimization

shaitan:
Fixed crash upon generating statistics for disabled at startup backends - parse disabled backends config.

Use first local address of node instead of empty at local_session - fixed "invalid address" in logs.

Connection between servers with different number of address is not allowed at the moment. Added corresponding check to dnet_process_socket to prevent appearing for short time node with different number of addresses in dht_state_list. This short time appearance leads to crash if someone's route list request is handled at the same time.

Optimized operations with groups list: underlying structure is replaced by rbtree.

### DIFF
--- a/library/dnet_common.c
+++ b/library/dnet_common.c
@@ -1,5 +1,6 @@
 /*
  * Copyright 2008+ Evgeniy Polyakov <zbr@ioremap.net>
+ * Copyright 2015+ Yandex
  *
  * This file is part of Elliptics.
  * 
@@ -620,6 +621,7 @@ int dnet_request_cmd(struct dnet_session *s, struct dnet_trans_control *ctl)
 	int num = 0;
 	struct dnet_net_state *st;
 	struct dnet_idc *idc;
+	struct rb_node *it;
 	struct dnet_group *g;
 	struct timeval start, end;
 	long diff;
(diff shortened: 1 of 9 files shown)


library/dnet_common.c
```diff
@@ -1,5 +1,6 @@
 /*
  * Copyright 2008+ Evgeniy Polyakov <zbr@ioremap.net>
+ * Copyright 2015+ Yandex
```
bioothod: Please do not add this, if you created file, add your own name

shaitan: It was made not in my commit, @toshic can you change it?

bioothod: This is a rather small feature compared to the whole file, just drop this.

I'm about to remove copyright notices and add authors file instead anyway. Authors have copyrights on corresponding code changes, which is observable via git tools

toshic: As mentioned before, this line was added by me. What's the problem with it?

bioothod: Copyrights are added to people for new file creation or major contribution, neither of it is present in the commit.

toshic: Is there any suggestions from FSF or any other foundation how to add copyright lines to files under GPL license? I found that copyright holders should be mentioned in notice.

bioothod: FSF assigns copyright to itself to prevent legal risks - it requires authors to yield appropriate rights at commit time.
I follow linux kernel rule of thumb - original author holds copyright notice and it can be extended when substantial part of the file is changed, in practice this only happens when more than half of the file is rewritten from scratch and it is not a plain refactor or improvement, but indeed a new subsystem.

Recently (like a couple of years) copyright tendency started to decline, authorship can be detected and shown via version control system, so Torvalds recommends to stop writing copyright notices at all.

shaitan:
Copyrights was removed.

shaitan:
@bioothod, can you merge these changes?

bioothod:
Sure, merged.

PR #580, opened 2015-03-27
Crash fix

abudnik:
More info see in commit descriptions


bioothod:
Comment description should look like this:

subsystem: short description

long description

Please use this style, it allows to easily generate changelog and quickly find what was affected

PR #582, opened 2015-04-02
build: fixed build on rhel 6 with boost v1.41.0-25

shaitan:


bioothod:
Does it work well with boost (can it be found, are all libraries linked and so on), i.e. what does this option mean?

shaitan:
I checked build on all our platforms. This option prevents FindBoost.cmake from picking up boost's BoostConfig.cmake.

bioothod:
We do not use boost config options and compile flags, so it should be safe to disable that options.
Patch applied, thank you

PR #585, opened 2015-04-08
network: send ping command to remote node if transactions stall count reaches its limit instead of resetting net_state

abudnik:
1. Problem description
If server node have multiple backends and one of this backend becomes suddenly very slow (due to some heavy operations like defrag), then client node reaches stall count limit of transactions, which were sended to this node. It leads to state resetting and all other transactions to healthy backends are freed. Server node becomes unreachable to client node, until reconnection process of client node setups connection to this server node.

2. Problem solution
Send non-blocking ping transaction (based on DNET_CMD_BACKEND_STATUS) to the stall node. If ping transaction timeouted, then do net_state reset.

Also this PR contains some refactoring around backend status requests in elliptics session.


library/trans.c
```diff
@@ -557,9 +557,88 @@ int dnet_trans_iterate_move_transaction(struct dnet_net_state *st, struct list_h
...
+static int dnet_ping_stall_node_complete(struct dnet_addr *addr __unused, struct dnet_cmd *cmd, void *priv)
+{
+	struct dnet_ping_node_private *ping_private;
+	struct dnet_net_state *st;
+
+	if (!is_trans_destroyed(cmd)) {
+	        ping_private = priv;
+		st = ping_private->net_state;
+
+		dnet_session_destroy(ping_private->session);
```
bioothod: This will crash if callback is being invoked multiple times.
You did not implement any reference counter on private data, why is it safe?

library/trans.c
```diff
@@ -557,9 +557,88 @@ int dnet_trans_iterate_move_transaction(struct dnet_net_state *st, struct list_h
...
+	ctl.data = &node_status;
+
+	ctl.complete = dnet_ping_stall_node_complete;
+	ctl.priv = ping_private;
+
+	err = dnet_trans_alloc_send_state(sess, st, &ctl);
+	if (!err)
+		goto err_out_exit;
+
+	dnet_session_destroy(sess);
```
bioothod: You destroy session which is passed to callback, that's a bug

library/trans.c
```diff
@@ -557,9 +557,88 @@ int dnet_trans_iterate_move_transaction(struct dnet_net_state *st, struct list_h
...
+
+	if (!is_trans_destroyed(cmd)) {
+	        ping_private = priv;
+		st = ping_private->net_state;
+
+		dnet_session_destroy(ping_private->session);
+		free(ping_private);
+
+		if (cmd->status == -ETIMEDOUT)
+			dnet_state_reset(st, cmd->status);
```
bioothod: If callback is invoked from dnet_session_destroy() called from dnet_ping_stall_node() this will deadlock, since dnet_ping_stall_node() is called under n->state_lock

library/trans.c
abudnik: >if callback is being invoked multiple times.

This guard condition prevents multiple time callback handling:
if (!is_trans_destroyed(cmd)) {

Also, I think that callback should not be called multiple times. Thereby is_trans_destroyed() should not exists at all.

library/trans.c
abudnik: Session destruction happens if dnet_trans_alloc_send_state() returns error code.
If dnet_trans_alloc_send_state() returns error code, then it wouldn't invoke any callback that passed to it as an argument (ctl.complete), so we must cleanup session and other resources hence.

library/trans.c
bioothod: Callback can and should be invoked multiple times - consider operations which are being split on server, or multiple replies (some data and status, or multiple data chunks like per-backend status and ack showing transaction has been completed). Your check says 'call this chunk when transaction is not about to be destroyed', which means potentially calling it multiple times.

You have removed dnet_update_status() function and its callback, which implemented correct reference counter mechanism which works for multiple replies. Your new code should use similar logic otherwise it is prone to crash.

library/trans.c
bioothod: dnet_trans_alloc_send_state() calls dnet_ping_stall_node_complete() on error.

Also, you changed error handling logic - it is always 'if error then go to error path', not vice versa, don't do that.

library/trans.c
abudnik: It shouldn't deadlock, because dnet_ping_stall_node_complete() is not called from dnet_ping_stall_node() directly.

bioothod: It is called directly when there is an error.

abudnik: Fixed problems above in commit: https://github.com/abudnik/elliptics/commit/d158b19dafa729fc0e9938c8169a94e151840a89

bioothod:
What about removing stall counter and instead use this new ping command to periodically check all remote nodes and reset connection if ping reply can not be received in time?

shaitan:
Pinging all remotes will kill net because it leads ping command to be send from all-to-all nodes. It is better to ping only remotes, connection with which has been compromised by stall counter.

bioothod:
Makes sence. I will review it again soon

PR #588, opened 2015-04-17
Check that backend supports iterator before run it.

shaitan:


bioothod:
Should it write error message to log in this case?

shaitan:
It will write error message right after `error_out_exit` label but with `DNET_LOG_NOTICE`. I can change the log level of the message in case of error.

bioothod:
But it will not tell us why -95 is returned, i.e. exactly who is responsible for this error

shaitan:
Added additional logs at iteration failures

PR #591, opened 2015-04-17
Added DNET_RECORD_FLAGS_UNCOMMITTED

shaitan:
* core: added common for all backends flag `DNET_RECORD_FLAGS_UNCOMMITTED` which is set for record is uncommitted and can't be read
 * iterator: made iterator collects uncommitted records
 * recovery: added handling uncommitted records, added options prepare-timeout which specifies timeout for uncommitted records after which such records should be deleted
 * recovery: fixed hanging on pool.close() - there is a bug in multiprocessing
 * pytests: added tests that checks new flag elliptics.record_flags.uncommitted:
   * serial of write_prepare, write_plain and write_commit with checking data accessibility
   * new tests of merge and dc recovery with mixing uncommitted keys into recovering keys.

### DIFF
--- a/include/elliptics/packet.h
+++ b/include/elliptics/packet.h
@@ -1009,8 +1009,21 @@ static inline void dnet_convert_iterator_request(struct dnet_iterator_request *r
 }
 
 
-/* Common record flags */
-#define DNET_RECORD_FLAGS_UNCOMMITTED	(1<<0) /* The record was prepared but hasn't been committed */
+/* eblob record flags */
+#define DNET_RECORD_FLAGS_REMOVE		(1<<0)
+#define DNET_RECORD_FLAGS_NOCSUM		(1<<1)
+#define DNET_RECORD_FLAGS_APPEND		(1<<4)
+/*
+ * This flag is set for records that are written in so-called extended format -
+ * records that have additional header before data - it's somewhat obscure and
+ * changes blob behaviour in various ways. Only user of this flag is elliptics.
+ */
+#define DNET_RECORD_FLAGS_EXTHDR		(1<<6)
+/*
+ * This flag is set for records that were prepared but haven't been commmitted yet.
+ * Such records doesn't available for read but can be committed even after restart.
+ */
+#define DNET_RECORD_FLAGS_UNCOMMITTED		(1<<7)
 
 /*
  * Iterator response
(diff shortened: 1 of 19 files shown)


shaitan:
Did you have time to review this pull request? Can it be merged?

include/elliptics/packet.h
```diff
@@ -1000,6 +1000,10 @@ static inline void dnet_convert_iterator_request(struct dnet_iterator_request *r
 	dnet_convert_time(&r->time_end);
 }
 
+
+/* Common record flags */
+#define DNET_RECORD_FLAGS_UNCOMMITTED	(1<<0) /* The record was prepared but hasn't been committed */
```
bioothod: Could we map this one-to-one to blob flags?
I think we might want to get other flags (like ext header or structured format) too someday

recovery/elliptics_recovery/types/merge.py
```diff
@@ -619,6 +624,7 @@ def recover(self):
         self.recover_result = Recovery(key=self.id,
                                        timestamp=self.timestamp,
                                        size=self.size,
+                                       flags=0,
```
bioothod: Why is it zero here?

include/elliptics/packet.h
shaitan: It means using eblob flags with all backends. Do we refuse independent backend realizations?

recovery/elliptics_recovery/types/merge.py
shaitan: This is method of `DumpRecover` that uses lookups for gathering meta of keys replicas. Lookup command doesn't provide record flags because `dnet_file_info` can't be changed without breaking elliptics protocol. Also for saving behaviour of reading/looking up uncommitted keys, these commands returns `-2` (`-ENOENT`) for such keys. This error can be changed to `-61` (`-ENODATA`) but still we can't provide records flags for recovering.

include/elliptics/packet.h
bioothod: 'uncommitted' is still a eblob-specific flag. We can try to map flags of the different backends to these new eblob-derived flags if needed, but I do not think this will be ever needed - for the last 6 years we had about 5-6 different backends and there were no external flags at all (and backends were removed eventually).

Having more eblob flags in iterator output and callback allows to check whether it has external header, if we will ever have structured objects stored, this flag will also help to check this.

shaitan: Ok, I just need an agreement that we end up with modular backends.

PR #592, opened 2015-04-17
memory leaks on reconnect failure due to timeout

abudnik:
socket->buffer was not released properly after failure in one of states in dnet_process_socket().

Refactored dnet_addr_socket & dnet_connect_state related resource handling logic.


library/net.cpp
```diff
@@ -63,27 +65,162 @@ enum dnet_socket_state {
...
+	dnet_connect_state *get()
+	{
+		atomic_add(&refcnt, 1);
+		return this;
+	}
+
+	void put()
+	{
+		if (atomic_dec_and_test(&refcnt))
+			delete this;
```
bioothod: Is this ever safe?

abudnik: dnet_connect_state_get, dnet_connect_state_put just moved into this class, so logically nothing has changed in dnet_connect_state ownership semantics (explicit refcnt'ing).

bioothod: I think deleting object from itself is undefined behaviour, is it?

abudnik: https://isocpp.org/wiki/faq/freestore-mgmt#delete-this

bioothod: Those set of rules basically says 'do not do this because it is very fragile'. It is exactly the reason unique and shared pointers exist as external containers.

Why is it ever needed to fix socket leak? You should split your patch into 2 parts: bugfix and this subtle refactoring.

If you decided to move away from manual c-like reference counters, you should use std::shared_ptr and not reinventing it from inside.

abudnik: std::shared_ptr is not available here, because dnet_connect_state owners are c-like structures (see dnet_request_route_list()).

Bugfix solution represented in refactoring code to cpp-like resource managment for dnet_socket_addr. Both dnet_socket_addr and dnet_connect_state changed their resource handling the same way.

And, again, nothing has semanticaly changed since last version. This implementation is safe and completely legal, even in sence of modern cpp memory managment.

euroelessar: What are the benefits of using dnet_connect_state::put function instead of dnet_connect_state_put? As I see first one only brings confusion and doesn't provide any benefits at all

abudnik:
This PR is a first iterat

library/net.cpp
abudnik: dnet_connect_state is not a POD-type right now (it contains stl container after changes in this PR). So, it is reasonable to handle its own resources destruction in destructor.
dnet_sonnect_state_put logic was decomposed into object lifetime control (put method) and object destruction (in d'tor).

abudnik:
No more "delete this;" : https://github.com/abudnik/elliptics/commit/fbd18ee095ae1d2780303dbba5262e92448394e6

library/net.cpp
```diff
@@ -494,116 +484,131 @@ static int dnet_validate_route_list(const char *server_addr, dnet_node *node, st
...
-	bool added_to_queue = false;
-	bool all_exist = false;
+		dnet_addr_container *cnt = reinterpret_cast<dnet_addr_container *>(cmd + 1);
+		const size_t states_num = cnt->addr_num / cnt->node_addr_num;
 
-	err = dnet_validate_route_list(server_addr, node, cmd);
-	if (err) {
-		goto err_out_exit;
-	}
+		std::unique_ptr<dnet_addr[]> addrs;
```
bioothod: Wow, can it be a vector?

library/net.cpp
```diff
@@ -612,27 +617,22 @@ void dnet_request_route_list(dnet_connect_state &state, dnet_net_state *st)
...
-			}
+		if (socket->s < 0) {
+			state->failed_count++;
+			continue;
 		}
 
+		const bool already_exist = state->sockets_connected.find(socket) != state->sockets_connected.end();
 		if (already_exist) {
-			dnet_log(state.node, DNET_LOG_NOTICE, "we are already connecting to %s",
+			dnet_log(state->node, DNET_LOG_NOTICE, "we are already connecting to %s",
```
bioothod: That should be 'we are already connected to'

library/net.cpp
```diff
@@ -792,29 +784,31 @@ static void dnet_process_socket(dnet_connect_state &state, epoll_event &ev)
...
-			state.node->indexes_shard_count = indexes_shard_count;
+			state->node->indexes_shard_count = indexes_shard_count;
 		}
 
-		socket->buffer = malloc(cmd->size);
-		socket->io_data = socket->buffer;
-		socket->io_size = cmd->size;
-
-		if (!socket->buffer) {
+		char *buffer = new(std::nothrow) char[cmd->size];
```
bioothod: What about making it std::vector instead of std::unique_ptr<char[]> ?

library/net.cpp
abudnik: https://github.com/abudnik/elliptics/commit/d17365b28822220aa276ef22ba52332375f26f9e

library/net.cpp
abudnik: https://github.com/abudnik/elliptics/commit/5748df8cbc60087cb1e7ce35c72e83bb84e23ed2

library/net.cpp
abudnik: I think it is not necessary here, because buffer is fixed-size and it doesn't need any extra vector semantics.

bioothod: Having any 'new' immediately rises concerns whether there are proper delete and so on. Even if they are correct, it requires more efforts still to check it, to find where data is allocated and freed and so on.

It is not a good practice to mix dynamic allocations and hidden allocations.

abudnik: https://github.com/abudnik/elliptics/commit/c6b275cb29d33130593fc110506d5f0fad9450ce

PR #593, opened 2015-04-21
network: use dht_state_num counter instead of iterating over list

abudnik:
We have 10-30% cpu constant load in dnet_node_state_num(), called from dnet_io_process_network()


bioothod:
I do not see any atomic_dec() which is definitely needed when state is removed from route table

library/net.c
```diff
@@ -1017,6 +1017,10 @@ int dnet_state_move_to_dht(struct dnet_net_state *st, struct dnet_addr *addrs, i
 	err = dnet_copy_addrs_nolock(st, addrs, addrs_count);
 
 	if (!err) {
+		list_for_each_entry(other, &st->node_entry, node_entry) {
+			atomic_inc(&st->n->dht_state_num);
```
bioothod: Why do you increase number of states by, well, number of connected states while you are actually adding only single state?

This whole math looks really bad in the patch

abudnik:
dht_state_list modified only in dnet_state_move_to_dht => it is not decreasing

library/net.c
abudnik: ok, will fix this

bioothod:
dht_state_list is being implicitly modified when state is removed. If state->node_entry lives in dht_state_list, then list_del(&state->node_entry) will remove entry from dht_state_list list

PR #594, opened 2015-04-22
network: use dht_state_num counter instead of iterating over list

abudnik:
Second try after #593


library/net.c
```diff
@@ -751,6 +751,9 @@ int dnet_process_recv(struct dnet_backend_io *backend, struct dnet_net_state *st
 
 void dnet_state_remove_nolock(struct dnet_net_state *st)
 {
+	if (st->__join_state == DNET_JOIN)
```
bioothod: JOIN state means this node joined others, i.e. this node had connected, sent JOIN request and auth command.
Client never sends this command, i.e. client never joins cluster.
Even more, when server accepts new client it can move it into dht list without JOIN state, i.e. when client sends JOIN command to given node.

What you need is a number of connected states for given **client** node.
This is a number of states in node->storage_state_list. Access to this list is a bit scattered over the code, but it can be easily refactored to single function. Please note list_move* functions operating this list too.
move operation has to increase number of states only if list_empty(&state->storage_state_entry) is true, i.e. given state has not been added yet.

bioothod: Even more, list_move*() for state->storage_state_entry should only be performed if state->storage_state_entry is empty, and it can be replaced with list_add_tail().

This can be done because state->storage_state_entry is supposed to be added only once and into single node->storage_state_list and not moved from client to dht list like dht_state_entry is being used.

abudnik: It was: list_for_each_entry(st, &n->dht_state_list, node_entry) {num++;

Why current implementation doesn't use node->storage_state_list?

bioothod: There was a single list previously. For client node it should contain the same nodes, since client can only connect to server nodes. While server node can connect to server nodes only, it can also accept client nodes. But only client is interested in number of connections and only interested in server connections.

Actually, I do not think this is still needed, since number of connections is very high these days. It was introduced when number of connections was around ten and we checked whether proxy is connected to at least half of them. With mastermind/backrunner management it is not needed.

Please talk about this number with colleagues, maybe we can remove it altogether. We do not use it in reverbrain proxies at all.

PR #595, opened 2015-04-24
don't use dnet_node_state_num() to estimate number of ready events in epoll

abudnik:
Third try after #593


bioothod:
Should it also remove this function and appropriate cpp binding?

abudnik:
session::state_num() used in proxy

PR #604, opened 2015-05-05
network: fixed bug with when total_count < failed_count in dnet_socket_c...

abudnik:
...onnect_new_sockets()

This bug inroduced in v2.26.5.0 (after net.cpp refactoring).


bioothod:
Please test your changes more thoroughly.
Can we add a test for reconnection/route processing which would catch this?

agodin:
Да, тест нужен.
Женя, подними версию, чтобы можно было протестить

bioothod:
There is 2.26.5.2 version in the repository

PR #605, opened 2015-05-08
Added elliptics network connectivity test

abudnik:
This PR contains test to bug, which was fixed in https://github.com/reverbrain/elliptics/pull/604

dnet_connectivity_test starts 20 nodes, where each single node sees only subset of all nodes, and after some time checks if a server node knows about all other nodes.


library/net.cpp
```diff
@@ -1100,6 +1100,8 @@ static int dnet_socket_connect(dnet_node *node, dnet_addr_socket_set &original_l
 	}
 
 	pthread_mutex_lock(&state->lock);
+
+	assert(state->succeed_count + state->failed_count <= state->total_count); // 'less' could be on timeout
```
bioothod: Does this mean application will crash without any error, without any notice, just out of nowhere?
That's no good - you should add a log at least, and better to handle this situation somehow - return an error, reset state, but not just blindly exit.

tests/connectivity_test.cpp
```diff
@@ -0,0 +1,168 @@
...
+	sleep(nodes_count);
+
+	register_tests(suite, sess);
+
+	return suite;
+}
+
+}
+
+int main(int argc, char *argv[])
```
bioothod: Please write a detailed description of the test - what does it do, what it expects, what will it do when something unexpected has being happened. Also specify commit it tests.

Btw, does this mean if I revert that commit, this test will fail? Did you test it that way?

library/net.cpp
euroelessar: Assert is effective only for debug builds, so it won't affect the production, only test environment.

bioothod: We always build with debug support. Maybe we do not set special compile-time definitions, but in this case it is useless - we do not make separate builds for tests, don't we?

tests/connectivity_test.cpp
abudnik: >Btw, does this mean if I revert that commit, this test will fail? Did you test it that way?

http://pastebin.com/kKTdv9iT

library/net.cpp
abudnik: >Does this mean application will crash without any error, without any notice, just out of nowhere?

It prints assertion to stdout and SIGABRT-ed (coredumped).

>we do not make separate builds for tests, don't we?

Yes, we don't.

tests/connectivity_test.cpp
bioothod: Test itself proved to be good, but test environment (asserts) is heavily flawed.
Please make it correct.

Also, for the further changes please make such a test first, see it failing and only then introduce the bugfix

library/net.cpp
abudnik: Also, I can write error message to logs before assert().

bioothod: You can not just crash the server because of some issues. Just can not. This is not allowed. If coredumps are turned off this ends up with spontaneous restarts which can not be debugged at all.

You have to write error message and mitigate error - either to reset connection or return error to client or whatever else, but not simply kill the server.

tests/connectivity_test.cpp
abudnik: This test did not always reproduce consequences of a race (when route list isn't converge), but always existence of this race itself.

library/net.cpp
abudnik: Some assert() statements are used in code base already.

tests/connectivity_test.cpp
bioothod: Why doesn't it reproduce this race reliably?
Can you change it and make more credible?

It is ok to have runtime checks in the code which end up with errors and subsequent actions (not crashes), this makes server bore robust, but test have to be reliable.

abudnik: We still don't know all conditions which make this bug always reproducible, but race condition catched pretty often.

library/net.cpp
bioothod: Those which I know of were made for impossible situations or cases where compilation issues do not allow server to work (ID size mismatch in eblob and elliptics and so on).

This is not about some bug we have just found, which can be triggered by client with different config option.

Basically, if this is something under remote control (remote config for instance), this is not allowed to crash.
If it is 110% local issue - this may assert with heavy error logs. But it has to be minimized.

tests/connectivity_test.cpp
bioothod: We have to know all conditions which lead to this bug. To any bug.
How can you fix a bug, if you do not know how it happens?

bioothod:
Test for races is not reliable, do not merge it now

PR #614, opened 2015-05-19
Recovery

shaitan:
* Moved gathering of route list into `get_routes()` at main process - made temporary node/session be freed after the route list is received.
* Allowed to use string value for '-L/--log-level'. Now log level can be specified by one of `debug`, `error`, `info`, `notice` or `warning`.

### DIFF
--- a/recovery/elliptics_recovery/recovery.py
+++ b/recovery/elliptics_recovery/recovery.py
@@ -123,7 +123,16 @@ def main(options, args):
 
     try:
         ctx.log_file = os.path.join(ctx.tmp_dir, options.elliptics_log)
-        ctx.log_level = int(options.elliptics_log_level)
+        try:
+            ctx.log_level = int(options.elliptics_log_level)
+        except:
+            ctx.log_level = options.elliptics_log_level
+
+        if isinstance(ctx.log_level, int):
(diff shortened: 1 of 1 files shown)


recovery/elliptics_recovery/recovery.py
```diff
@@ -55,6 +55,25 @@ def cleanup(path):
...
 
 
+def get_routes(ctx):
+    log.debug('Requesting routes')
+
+    log.debug("Creating logger")
+    elog = elliptics.Logger(ctx.log_file, int(ctx.log_level))
+
+    log.debug("Creating node")
+    node = elliptics_create_node(address=ctx.address,
```
bioothod: Why does it need temporal node/session?
Could it be better to use the same node to get routes and process all requests?

shaitan: All requests are sent from subprocesses and, currently, there is no possibility to pass initialized node from main process to subprocesses.

bioothod: Could those processes grab route table from own nodes?

Or route information is used to split requests among processes?

shaitan: Yes, main process splits task between subprocesses based on route list.

PR #615, opened 2015-05-22
eblob: when using prepare+plain_write+commit flags, commit number of …

bioothod:
…written bytes, not number of allocated bytes on disk


shaitan:
New behaviour of write with prepare+plain_write+commit flags is not obvious and should be described in detail in the documentation. Also it should be covered by post-build test.

bioothod:
This is actually a bugfix of the case which accidentally was not yet used by others.
Currently setting up all 3 flags ends up committing a bunch of zeroes into the data, this is not how I designed prepare/commit methods.

shaitan:
Your design of prepare/plain_write/commit methods should be presented in documentation and covered by tests.

PR #616, opened 2015-05-22
network: sockets are lost from node's reconnect list after dropped co…

abudnik:
…nnection to net state (after stall_count timeout)


bioothod:
I thought there was a test for reconnection logic?
Can we create one with show client-side wait timeouts (like a second) and large enough backend delay.
This should force client timeout and stall counter to increase, which will eventually end up with connection reset and reconnection. The latter can be checked by reading route table

bioothod:
Test is pending, will be in the next release

PR #617, opened 2015-05-27
Fixes and optimizations

shaitan:
Fixed small memory leak of `dnet_config_backend` - call cleanup.
Used const pointers for dnet_config_entry callback arguments and removed useless copying of value.

### DIFF
--- a/example/eblob_backend.c
+++ b/example/eblob_backend.c
@@ -1147,10 +1147,25 @@ err_out_exit:
 	return err;
 }
 
+
+/*
+ * dnet_blob_config_cleanup() stops and cleans up eblob if it is needed and
+ * frees memory allocated by config.
+ * There are 3 stages of backend:
+ *   1. config parsing
+ *   2. eblob initialization
+ *   3. cleaning up eblob and config
+ * In common case backend goes throught all 3 steps.
+ * But there is specific case (backends_stat_provider.cpp: @fill_disabled_backend_config()):
+ *   monitor subsystem for disabled backends makes only 1 and 3 stages and skips stage 2.
+ *   So when monitor makes stage 3, it has uninitialized eblob and
+ *   should cleanus up only config data.
+ */
 static void dnet_blob_config_cleanup(struct dnet_config_backend *b)
 {
 	struct eblob_backend_config *c = b->data;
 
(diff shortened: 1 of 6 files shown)


example/eblob_backend.c
```diff
@@ -976,7 +988,9 @@ static void eblob_backend_cleanup(void *priv)
 {
 	struct eblob_backend_config *c = priv;
 
-	eblob_cleanup(c->eblob);
+	if (c->eblob) {
```
bioothod: Can we call eblob_backend_cleanup() without eblob?

shaitan: There are 2 way to cleanup backend:
  1. by calling `backend_cleanup` from `dnet_backend_callbacks` - unused and can be removed
  2. by calling `cleanup` from `dnet_config_backend`

The last one can be called even if backend is uninitialized for freeing resource allocated while parsing config.

bioothod: The only place where we initialize locks and data is dnet_blob_config_init(), this is also the only place where we initialize $eblob field. How is it possible to call cleanup() with half initialized backend config structure?

Also, it looks like cleanup can not be called without proper backend config initialization at all. Since it is exported to c++ world only via init/cleanup callbacks.

shaitan: I've moved checking `c->eblob` intp `dnet_blob_config_cleanup` and now it will free only `c->data.file` if blob wasn't initialized.

example/eblob_backend.c
```diff
@@ -1140,7 +1151,10 @@ static void dnet_blob_config_cleanup(struct dnet_config_backend *b)
 {
 	struct eblob_backend_config *c = b->data;
 
-	eblob_backend_cleanup(c);
+	if (c->eblob)
+		eblob_backend_cleanup(c);
+
+	free(c->data.file);
 }
```
bioothod: This requires a long description of how is it possible, that we have half-filled control structure, but did not initialize eblob. Including two-steps init process (parse config, then call ->init()), and that parsing disabled backend configs will not have second stage, but we will have to cleanup resources allocated during config parsing.

shaitan: I've added description to `dnet_blob_config_cleanup()` and put details for monitor case.

PR #629, opened 2015-06-29
Added support of new eblob checksums

shaitan:
Support changes from https://github.com/reverbrain/eblob/pull/131.

### DIFF
--- a/include/elliptics/packet.h
+++ b/include/elliptics/packet.h
@@ -747,7 +747,9 @@ struct dnet_io_attr
 	 */
 	uint64_t		total_size;
 
-	uint64_t		reserved1;
+	/* Combination of DNET_RECORD_FLAGS_* */
+	uint64_t		record_flags;
+
 	uint32_t		reserved2;
 
 	uint32_t		flags;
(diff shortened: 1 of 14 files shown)


agend:
Will it be complicated to migrate to new major version?

shaitan:
No, it will not be. There is no need to make any special steps.
But it will be one-way migration in the sense that the previous version will fail to verify checksums of entries written after migration. To read such entries by the previous version, client should disable checksum verification by setting `DNET_IO_FLAGS_NOCSUM`. Entries written before migration will remain unmodified and could be verified by both versions.

include/elliptics/packet.h
```diff
@@ -747,7 +747,9 @@ struct dnet_io_attr
 	 */
 	uint64_t		total_size;
 
-	uint64_t		reserved1;
+	/* Combination of DNET_RECORD_FLAGS_* */
+	uint64_t		record_flags;
```
bioothod: How does it differ from io->flags? There will be a serious intersection between the two - we should implement it the way that does not introduce duplication.

shaitan: It is completely different from `io->flags` because `io->flags` is about how concrete read/write/remove command should be done, but `record_flags` is about how the entry is stored in the blob.

PR #632, opened 2015-07-09
oplocks: use dnet_id instead of dnet_raw_id for taking into account group_id

abudnik:


library/locks.c
```diff
@@ -103,6 +103,14 @@ int dnet_locks_init(struct dnet_node *n, int num)
 	return err;
 }
 
+static inline int dnet_locks_entry_compare(const struct dnet_locks_entry *entry, const struct dnet_id *id)
```
bioothod: Please add comment about why we decided to switch from raw ID comparison to raw ID + group ID comparison.
I.e. that we want the same keys in different groups not to interfere with each other

abudnik: done

PR #636, opened 2015-07-20
New oplocks: get request from queue, whose key id is not already locked

abudnik:
dnet_io_process() calls dnet_take_request() to take request from dnet_work_pool's queue. It skips already locked keys while searching linearly in request queue.

Load test: write keys at 10k rps, where 50% of all keys are the same (with vm.dirty_background_ratio=0). Load test results:
[Active threads (new oplock)](https://files.slack.com/files-pri/T04E9PMS0-F093NAG7M/download/active_threads_new.png) vs. [Active threads (current oplock)](https://files.slack.com/files-pri/T04E9PMS0-F093NN4AZ/download/active_threads_old.png)
[Quantilies (new oplock)](https://files.slack.com/files-pri/T04E9PMS0-F093NGNSJ/download/quantilies_new.png) vs. [Quantilies (current oplock)](https://files.slack.com/files-pri/T04E9PMS0-F093NB5QC/download/quantilies_old.png)

50% quantile of requests are responded in less than 85ms, while other requests responded in 2 seconds (new oplocks implementation). Half of requests with different keys are processed fast, while other half of requests with same key formed a queue.
In old oplock implementation every request processed in 2 seconds, because their processing time bounded by time of waiting in oplocks for requests with same key.
Also average number of active threads shows that i/o threads are not waiting in oplocks.

New oplock implementation changes elliptics write order semantics of requests with same key: multiple writes to a single backend will be processed in a strict order as they came from client. Older (current) oplock implementation doesn't guarantee processing order of requests with same key.


library/request_queue.h
```diff
@@ -0,0 +1,96 @@
...
+{
+	return !dnet_id_cmp(&lhs, &rhs);
+}
+
+struct dnet_locks_entry
+{
+	std::condition_variable unlock_event;
+};
+
+class dnet_request_queue
```
bioothod: You absolutely **have** to write a detailed description of the new locking design.
This is a substantial part of the processing engine, and I've already spotted issues with mutexes.

library/request_queue.h
```diff
@@ -0,0 +1,96 @@
...
+	void put_lock_entry(dnet_locks_entry *entry);
+
+private:
+	struct list_head m_queue;
+	std::mutex m_queue_mutex;
+	std::condition_variable m_queue_wait;
+
+	std::atomic_ullong m_queue_size;
+
+	std::unordered_map<dnet_id, dnet_locks_entry *> m_locked_keys;
```
bioothod: This is the slowest 'sorted' container ever.

Although this should be a fair hash table with constant access time, the last time I checked performance of the implementation it was several times slower than std::map for search and insertion.
It would be a good idea to run performance test for common containers.

library/dnet.c
```diff
@@ -761,14 +761,6 @@ static int dnet_cmd_bulk_read(struct dnet_backend_io *backend, struct dnet_net_s
 		cmd->flags &= ~DNET_FLAGS_NEED_ACK;
 	}
 
-	/*
-	 * we have to drop io lock, otherwise it will be grabbed again in dnet_process_cmd_raw() being recursively called
-	 * Lock will be taken again after loop has been finished
-	 */
-	if (!(cmd->flags & DNET_FLAGS_NOLOCK)) {
-		dnet_opunlock(st->n, &cmd->id);
```
shaitan: Why did you remove it? `cmd->id` coincides with one of ids from ios, so it should be unlocked before calling `dnet_process_cmd_raw` and should be locked after.

shaitan: Do we have bulk operation tests? If no, please add them because this looks buggy.

abudnik: >Do we have bulk operation tests?
We have such test in test.cpp, but bulk read tests as any 'read' operation tests always successfull in context of oplocks, because read operation use send_file()

abudnik: >Why did you remove it?

I will add oplocks as well as it used in elliptics cache sync code.

example/monitoring-stats.txt
```diff
@@ -6,10 +6,6 @@ io.cmd
 io.cmd.{<any elliptics command type in upper case>}
     time of commands for each type
 
-io.cmd.{<any elliptics command type in upper case>}.lock_time
-    time spent in id lock waiting, for each command type,
-    (this will be 0 for commands with DNET_FLAGS_NOLOCK flag set)
-
```
bioothod: Why do you remove this? Isn't it still possible to wait on locked command if there are only already locked keys in the queue?

library/dnet.c
```diff
@@ -1119,9 +1112,6 @@ int dnet_process_cmd_raw(struct dnet_backend_io *backend, struct dnet_net_state
 
 	err = dnet_send_ack(st, cmd, err, recursive);
 
-	if (!(cmd->flags & DNET_FLAGS_NOLOCK))
-		dnet_opunlock(n, &cmd->id);
-
```
bioothod: Wow, you've just removed unlock for every command, that is a bug

abudnik: Any command processed by dnet_process_cmd_raw() came from i/o thread pool reactor: dnet_io_process(). dnet_io_process() use dnet_pop_request() to take *always* non locked key.

example/monitoring-stats.txt
abudnik: If there are only already locked keys in the queue, then i/o pool threads will wait for conditional variable (dnet_request_queue::m_queue_wait), but they will not wait for any lock.

library/request_queue.cpp
```diff
@@ -0,0 +1,288 @@
...
+	if (!(cmd->flags & DNET_FLAGS_REPLY) &&
+	    !(cmd->flags & DNET_FLAGS_NOLOCK)) {
+		release_key(&cmd->id);
+	}
+}
+
+void dnet_request_queue::lock_key(const dnet_id *id)
+{
+	std::unique_lock<std::mutex> lock(m_locks_mutex);
+	while (1) {
```
bioothod: Does it mean, that lock can loop forever and miss `need_exit` flag set by signal for example?

abudnik: Key will be unlocked even if need_exit is set, therefore this loop is finite.

bioothod: So, we have the following order:
1. signal/client sets `need_exit`
2. exiting routing detects this flags and kills all keys are requests
3. this thread sees, that there is no key
4. it adds new key into lock hash table
5. it performs its job
6. releases key
7. exits since it sees `need_exit` at higher level

Is that correct?

abudnik: while(!need_exit) {
r=dnet_pop_request() // locks key
dnet_process_recv(r)
dnet_release_request(r) // unlocks key
}

bioothod:
Looks good to me, thank you
I will merge this

PR #637, opened 2015-07-21
eblob: fixed integer overflow for uncommitted records during iteration

bioothod:


example/eblob_backend.c
```diff
@@ -127,7 +127,18 @@ static int blob_iterate_callback_common(struct eblob_disk_control *dc, int fd, u
...
+		/*
+		 * When record is not yet committed (it was allocated on disk via write_prepare and optional write_plain),
+		 * its @data_size is zero and removing this header size ends up with negative size converted back to
+		 * very large positive number (0xffffffffffffffd0)
+		 *
+		 * @data_offset above is safe, since actually eblob had allocated all needed space,
+		 * it just hasn't yet filled eblob_disk_control
+		 */
+		if (!(dc->flags & BLOB_DISK_CTL_UNCOMMITTED)) {
+			size -= sizeof(struct dnet_ext_list_hdr);
```
shaitan: Elliptics shoudn't set `BLOB_DISK_CTL_EXTHDR` to the record flags until extended header is written.

bioothod: Elliptics sets BLOB_DISK_CTL_EXTHDR at 'prepare' time, writing external header is exactly the same as writing other data: https://github.com/reverbrain/elliptics/blob/master/example/eblob_backend.c#L177

This allows iterator to check this record before any data including ext header has been yet written (@data_size equals zero), but record already exists on disk

shaitan: Records without ext header should not have `BLOB_DISK_CTL_EXTHDR` in its flags because it is wrong, so Elliptics shouldn't set this flag at 'prepare' time. I suggest to set this flag only at 'pain_write' or 'commit' time when Elliptics really writes ext header.

bioothod: Makes sence, I will accept this pull request since it fixes bug with already stored keys, and will cook up the next one with setting ext header flag only when ext header is being written

shaitan: This pull request breaks post-built recovery test.
**Initial state**:
 * First group hasn't key
 * Second group has prepared but uncommitted key with newest timestamp
 * Third group has committed key with old timestamp

**Expected result after recovery**: state of the key in all three groups wasn't changed .
**Got:** reading the key from first group is succeeded, so recovery wrote it.

bioothod: Apparently this test will (or has to) fail without the change too. If it doesn't, your test is wrong and has to be fixed.
This commit doesn't change logic of the key reading, it only changes the size reported by the iterator.

This change fixes only size of the entry to be reported during iteration time like this one:
node: address:port, key: 1ae593e6c54eeb97c0b9590e980afd500bc0f8352bf6aef60e671c093413c315c8950c3ba3a60e6b231ba7a792a13e3b2b22f052f3c8ef64352ceda29240479f, flags: 0, ts: 1437326697/750386000, keys: 70366/95987, status: 0, size: 18446744073709551568, data:

18446744073709551568 size is 0xffffffffffffffd0 - or minus sizeof(struct dnet_ext_list_hdr)

shaitan: Or there is a recovery bug revealed by the changes.

shaitan: Yep, it skips iterator response with size == 0. So in that case recovery will drop info about such replicas and overwrites just prepared replicas by old version.

To get correct behaviour:
 * Elliptics should write ext header at any write: 'prepare', 'plain', 'commit' and regular write.
 * recovery shouldn't skips iterator response with zero data size

bioothod: I've added ext header write for prepare/commit phases, take a look: https://github.com/reverbrain/elliptics/pull/637/files#diff-5a41c5529883101effa307ea0e7ca3fdL200

Although this changes on-disk logic - previously we only wrote ext header when there is actual data, but this is not valid. Check is still needed for old data.

shaitan: I agree, but isn't it better to check the size before reading ext header? If old data was written without ext header what header will we read?

shaitan: Changes in `iterate.cpp` can't be compiled at lucid:
```
/tmp/elliptics-2.26.6.1/example/iterate.cpp: In function 'int main(int, char**)':
/tmp/elliptics-2.26.6.1/example/iterate.cpp:171: error: 'class boost::program_options::typed_value<std::vector<std::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::allocator<std::basic_string<char, std::char_traits<char>, std::allocator<char> > > >, char>' has no member named 'required'
```

Do we support lucid?

shaitan: Also last changes broke post-build tests:
```
/tmp/buildd/elliptics-2.26.6.1/tests/test.cpp(549): fatal error in "(test_prepare_commit, create_session(n, {1, 2}, 0, 0), "prepare-commit-test-1", 0, 0)": (sess.write_prepare(remote, prepare_data, offset, total_size_to_reserve)), err: "1:2d901b550a02...3a8a54402c72: Failed to process WRITE command: Operation not supported: -95"
/tmp/buildd/elliptics-2.26.6.1/tests/test.cpp(549): fatal error in "(test_prepare_commit, create_session(n, {1, 2}, 0, 0), "prepare-commit-test-2", 0, 1)": (sess.write_prepare(remote, prepare_data, offset, total_size_to_reserve)), err: "1:4a780fd8b60b...a2dc4bc6a0f6: Failed to process WRITE command: Operation not supported: -95"
/tmp/buildd/elliptics-2.26.6.1/tests/test.cpp(549): fatal error in "(test_prepare_commit, create_session(n, {1, 2}, 0, 0), "prepare-commit-test-3", 1, 0)": (sess.write_prepare(remote, prepare_data, offset, total_size_to_reserve)), err: "1:bbe6a135940c...3c5ca5c872cb: Failed to process WRITE command: Operation not supported: -95"
/tmp/buildd/elliptics-2.26.6.1/tests/test.cpp(549): fatal error in "(test_prepare_commit, create_session(n, {1, 2}, 0, 0), "prepare-commit-test-4", 1, 1)": (sess.write_prepare(remote, prepare_data, offset, total_size_to_reserve)), err: "1:a8d79203b72c...499cf27c6801: Failed to process WRITE command: Operation not supported: -95"
/tmp/buildd/elliptics-2.26.6.1/tests/test.cpp(464): fatal error in "(test_prepare_commit_simultaneously, create_session(n, {1, 2}, 0, 0))": (s.write_data(ctl)), err: "1:8225dd40be99...45504df808e3: Failed to process WRITE command: Operation not supported: -95"
```

bioothod: Data is always being written with ext header. Problem was in @flags which were set at prepare time, while ext header and data had been written later at plain_write phase.

I agree, that it is not safe just to check flags, since we introduce another bug - perfectly valid but yet uncommitted record does have an ext header, but also BLOB_DISK_CTL_UNCOMMITTED flag. This will end up with incorrect situation when size is not reduced. I've updated patch.

bioothod:
I've updated code and comments section.
It is not allowed not to set ext header bit at prepare time, it must be set otherwise overwrite will not distinguish this record from old records without ext header at all, and will not allow to overwrite, thus -95 (-ENOTSUP) error.

Alternative way to fix it is to change eblob's eblob_plain_writev_prepare() check, but it is hard to make right. I also updated comments about these cases.

example/iterate.cpp
```diff
@@ -169,13 +167,15 @@ int main(int argc, char *argv[]) {
 	desc.add_options()
 	("group,g", boost::program_options::value<std::vector<int>>()->multitoken(), "group IDs to connect")
 	("log-file,l", boost::program_options::value<std::string>()->default_value("/dev/stderr"), "log file")
-	("log-level,L", boost::program_options::value<int>()->default_value(1), "log level")
-	("remote,r", boost::program_options::value<std::vector<std::string>>()->multitoken(), "adds a route to the given node")
+	("log-level,L", boost::program_options::value<std::string>()->default_value("info"), "log level")
+	("remote,r", boost::program_options::value<std::vector<std::string>>()->multitoken()->required(), "adds a route to the given node")
```
bioothod: As of lucid, I would drop its support, do you have any servers that old?
I think lucid doesn't even have c++11 support which is required for elliptics build

shaitan:
:+1: Looks good for me, thank you.

PR #645, opened 2015-08-31
crypto: fixed invalid processing of data block larger than buffer in …

abudnik:
…sha512_file()/sha512_file_ctx()


abudnik:
ping

PR #646, opened 2015-09-07
localnode access service

ijon:
DO NOT MERGE

This PR contains work on `localnode` service.

`localnode` service is a cocaine service spinned off by the srw and its purpose is to give srw workers and other local (also remote) users easy access to the facilities of the local elliptics node.
`localnode` closely resembles existing `elliptics` service, but unlike it `localnode` directly communicates with server node, so there is no middle layers, no additional costs and no configuration.
That is the main reason behind `localnode`.

This code is actually a revival of someone's previous effort, dusted off and cleaned up a bit.

As it is, `localnode` service is in working condition and is tested, but it's not ready for inclusion into a mainline yet.

Future direction:
- unification with `elliptics` service, they are supposedly too similar to have them both; it could be one implementation (interface layer and all) with initialization which takes either client elliptics node (`elliptics` service in a pure cocaine cloud) or server elliptics node (`localnode` running inside srw)
- interface evolution, right now its sketchy
- close fitting for use by and with srw workers (exec support, sending replies etc.)
- texts cleanup
##### Ugliness

`cocaine/include/cocaine/actor.hpp` is a pristine copy of actor.hpp from cocaine-core sources. There is no other way, unfortunately: srw require actor_t defined in actor.hpp to be able to register localnode without requiring external config, but, by whatever reason, actor.hpp doesn't considered a part of public interface in cocaine v0.11 (in cocaine v0.12 it does).

---
This change is [Reviewable](https://reviewable.io/reviews/reverbrain/elliptics/646)

### DIFF
--- a/tests/srw_test.cpp
+++ b/tests/srw_test.cpp
@@ -13,19 +13,86 @@
  * GNU Lesser General Public License for more details.
  */
 
-#include "srw_test.hpp"
-#include "test_base.hpp"
 #include <algorithm>
 
+#include <boost/program_options.hpp>
 #define BOOST_TEST_NO_MAIN
 #include <boost/test/included/unit_test.hpp>
 
-#include <boost/program_options.hpp>
+#include "elliptics/interface.h"
+
+// force cocaine templates to choose std's placeholders over boost's
+namespace cocaine { namespace io {
+	using std::placeholders::_1;
+	using std::placeholders::_2;
+}}
+#include "cocaine/framework/services/localnode.hpp"
+
+#include "test_base.hpp"
+#include "srw_test.hpp"
+
+namespace {
+
+std::string to_hex_string(const unsigned char * data, const size_t size)
+{
+	static const char hexchars[] = "0123456789abcdef";
+	std::string result;
+	result.reserve(size * 2);
+	for (size_t i = 0; i < size; ++i) {
+		const unsigned char byte = data[i];
+		result.push_back(hexchars[(byte >> 4) & 0x0f]);
+		result.push_back(hexchars[byte & 0x0f]);
+	}
+	return result;
+}
+
+std::string gen_random(const int len) {
+	static const char alphanum[] =
+		"0123456789"
+		"ABCDEFGHIJKLMNOPQRSTUVWXYZ"
+		"abcdefghijklmnopqrstuvwxyz";
+
+	std::string result;
+	result.reserve(len);
+	for (int i = 0; i < len; ++i) {
+		result += alphanum[rand() % (sizeof(alphanum) - 1)];
+	}
+	return result;
+}
+
+}
+
+// required for localnode_test
+//
+namespace ioremap { namespace elliptics {
+
+inline bool operator==(const dnet_async_service_result &a, const dnet_async_service_result &b)
+{
+	return ((0 == dnet_addr_cmp(&a.addr, &b.addr))
+		&& (0 == memcmp(&a.file_info, &b.file_info, sizeof(a.file_info)))
+		&& a.file_path == b.file_path
+	);
+}
+
+inline std::ostream& operator<<(std::ostream& ostr, const dnet_async_service_result &v)
+{
+	ostr << "{\n addr=" << dnet_addr_string(&v.addr)
+		<< ",\n " << "record_flags=" << v.file_info.record_flags
+		<< ",\n " << "size=" << v.file_info.size
+		<< ",\n " << "time=" << dnet_print_time(&v.file_info.mtime)
+		<< ",\n " << "file=" << v.file_path
+		<< ",\n " << "checksum=" << to_hex_string(v.file_info.checksum, sizeof(v.file_info.checksum))
+		<< "\n}"
+		;
+	return ostr;
+}
+
+}}
+
+namespace tests {
 
 using namespace ioremap::elliptics;
 using namespace boost::unit_test;
-
-namespace tests {
 
 static std::shared_ptr<nodes_data> global_data;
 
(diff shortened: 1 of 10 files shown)


srw/direct_access.cpp
```diff
@@ -0,0 +1,205 @@
...
+	, m_session_proto(node)
+{
+	COCAINE_LOG_INFO((&m_log), "%s: enter", __func__);
+
+	on<direct_access::read_data>("read_data", std::bind(&direct_access_t::read_data, this, _1, _2, _3));
+	on<direct_access::lookup>("lookup", std::bind(&direct_access_t::lookup, this, _1));
+	on<direct_access::write_data>("write_data", std::bind(&direct_access_t::write_data, this, _1, _2, _3));
+
+	// find all groups that resides on our node, and set them as session default
+	m_session_proto.set_groups(find_local_groups(m_session_proto));
```
bioothod: This will force `m_session_proto` to perform action on every single group which is present in current server's config.
It is probably not intended behaviour

srw/direct_access.cpp
```diff
@@ -0,0 +1,205 @@
+/*
+ * 2014+ Copyright (c) Yandex
```
bioothod: Not Yandex, but Ivan Chelubeev.
Copyright belongs to authors In Elliptics.

tests/srw_test.cpp
```diff
@@ -160,6 +213,143 @@ static void timeout_test(session &sess, const std::string &app_name)
...
+		BOOST_REQUIRE_EQUAL(deferred.valid(), true);
+		BOOST_CHECK_GT(result.file_info.size, 0);
+		BOOST_CHECK_GT(result.file_path.size(), 0);
+	}
+
+	//XXX: can't compare write_result and lookup_result directly because
+	// file_info.size meaning and value differs for write and lookup operations
+	//BOOST_CHECK_EQUAL(write_result, lookup_result);
+	{
+		BOOST_CHECK_EQUAL(std::string((const char *)&write_result.addr, sizeof(dnet_addr)), std::string((const char *)&write_result.addr, sizeof(dnet_addr)));
```
shaitan: Should it compare `write_result.addr` and `lookup_result.addr`? Currently, it compares `write_result.addr` with itself.

tests/srw_test.cpp
```diff
@@ -160,6 +213,143 @@ static void timeout_test(session &sess, const std::string &app_name)
...
+		BOOST_CHECK_EQUAL(write_result.file_path, lookup_result.file_path);
+#define CMP(x) BOOST_CHECK_EQUAL(write_result.file_info.x, write_result.file_info.x)
```
shaitan: The same: it compares `write_result.file_info.x` with itself.

tests/srw_test.cpp
```diff
@@ -160,6 +213,143 @@ static void timeout_test(session &sess, const std::string &app_name)
...
+	auto value = gen_random(15);
+
+	dnet_async_service_result write_result;
+	dnet_async_service_result lookup_result;
+	{
+		auto &result = write_result;
+		auto deferred = localnode->write_data(key.id(), value, 0);
+		BOOST_REQUIRE_NO_THROW(result = deferred.next());
+		BOOST_REQUIRE_EQUAL(deferred.valid(), true);
+		BOOST_CHECK_GT(result.file_info.size, 0);
```
shaitan: Why doesn't it check that size is equal to size of original value?

srw/direct_access.cpp
```diff
@@ -0,0 +1,205 @@
...
+deferred<dnet_async_service_result> direct_access_t::write_data(const dnet_id &id, const std::string &data, uint64_t offset)
+{
+	COCAINE_LOG_INFO((&m_log), "%s: enter", __func__);
+
+	deferred<dnet_async_service_result> promise;
+
+	auto s = m_session_proto.clone();
+	s.set_exceptions_policy(ioremap::elliptics::session::no_exceptions);
+
+	if (auto error = override_groups(s, id.group_id)) {
```
shaitan: There should exist a way to write/lookup/read data to/from a set of groups.

srw/direct_access.cpp
```diff
@@ -0,0 +1,205 @@
...
+		const ioremap::elliptics::error_info &error)
+{
+	COCAINE_LOG_INFO((&m_log), "%s: enter", __func__);
+
+	if (error) {
+		COCAINE_LOG_INFO((&m_log), "%s: return error %d, %s", __func__, error.code(), error.message());
+		promise.abort(error.code(), error.message());
+	} else {
+		dnet_async_service_result r;
+		const auto &single_result = results[0];
```
shaitan: Write can have several results and all of them should be returned to worker.

tests/srw_test.cpp
ijon: Indeed.

tests/srw_test.cpp
ijon: Because `write` and `lookup` commands return different sizes -- `lookup` returns user data size while `write` return user data size plus some extra (some header size, may be). Or vice versa.

srw/direct_access.cpp
ijon: Group list in session_proto work only as default, it gets corrected on every operation.

srw/direct_access.cpp
ijon: Module comment blocks just hadn't been cleaned up at the time. Now they are done.
But I'm not comfortable with putting personal copyright strings in source files (those names and dates just add clutter in my opinion and also become obsolete quickly).
Maybe it is finally a time to have an AUTHORS file (as you proposed some time ago)?

srw/direct_access.cpp
ijon: Did exactly that.

PR #650, opened 2015-09-14
Server-send operations

bioothod:
This pull request introduces 2 server-side operations which allow to send content from one server to multiple replicas. It is intended for various recovery tools which optimize by not copying data from local node to recovery temporary location, instead they may tell remote node to send data directly to required locations. It can also be used to move data from older eblob version to newer without server interruption.

1. New iterator type which sends all keys being iterated to set of remote groups. It does it with the speed of network or disk (what it slower), in local tests iteration over 200Gb blobs sending data over the network to one remote node ended up with ~78MB/s sustained speed with pikes though, especially when remote node synced caches. Both sender and recipient had 30Gb of RAM. `rsync` shows ~32MB/s speed on these machines, but not because it is that slow, but because of `ssh` which maxed out CPU by packet encryption.
Iterator sends `dnet_iterator_response` structure for each write result for every key it has processed just like for usual iterator, neither API nor ABI is broken.

2. Server-send command which accepts vector of keys. API searches for all remote nodes/backends which host given keys in the one specified group, splits keys into per-node/backend basis and tells remote backends to send appropriate keys to specified remote groups. The same iterator response is generated for every key which has been processed.

There are 3 operation modes:
1. default - writing data to remote node using compare-and-swap, data will only be written if it doesn't exist on remote servers. Server send (iterator or per-key) running in this mode is especially useful for recovery - there is no way it can overwrite newer copy with the old data.
2. overwrite - when special flag is set, overwrite data (clears compare-and-swap logic)
3. move - if write has been successful, remove local key

There is example tool in example dir, which works the same way as `example/iterate.cpp`, but performs copy/move of the iterator results. There are 16 various tests which cover all operation modes. Tests are rather long - so far they are the longest, it takes about 170 seconds on my server.

Unfortunately this pull request also contains fair number of bugfixes which should live in a separate commits. These include possible memory corruption, invalid pointer dereference and race conditions. I found them during extensive testing and failed to separate into own commits.


bindings/cpp/logger.cpp
```diff
@@ -137,13 +136,12 @@ void dnet_node_set_trace_id(dnet_logger *logger, uint64_t trace_id, int tracebit
 	if (scoped_count >= DNET_SCOPED_LIMIT) {
 		dnet_log_only_log(logger, DNET_LOG_ERROR,
 			"logic error: you must not call dnet_node_set_trace_id twice, dnet_node_unset_trace_id call missed");
-		++scoped_count;
+		// crash
+		char *ptr = NULL;
+		*ptr = 1;
```
shaitan: What is this crash designed for?

bindings/cpp/logger.cpp
```diff
@@ -156,32 +154,35 @@ void dnet_node_set_trace_id(dnet_logger *logger, uint64_t trace_id, int tracebit
 			attributes.insert(std::make_pair(std::string("backend_id"), blackhole::log::attribute_t(backend_id)));
 		}
 
-		new (local_attributes) scoped_attributes_t(*logger, std::move(attributes));
+		scoped_attributes[scoped_count] = new scoped_attributes_t(*logger, std::move(attributes));
```
shaitan: If you remove `scoped_buffer`, there is no profit to use `blackhole::scoped_attributes_t *scoped_attributes[DNET_SCOPED_LIMIT]` and it is better to replace it by `std::vector<blackhole::scoped_attributes_t >`.

bindings/cpp/logger.cpp
```diff
@@ -156,32 +154,35 @@ void dnet_node_set_trace_id(dnet_logger *logger, uint64_t trace_id, int tracebit
...
-
-	++scoped_count;
 }
 
 void dnet_node_unset_trace_id()
 {
 	using namespace blackhole_scoped_attributes;
 
-	--scoped_count;
+	if (scoped_count > 0) {
```
shaitan: `scoped_count` should not be zero here, so if it is, there is a bug and `dnet_node_unset_trace_id()` was called too many times.

bindings/cpp/session.cpp
```diff
@@ -2233,12 +2248,14 @@ error_info session::mix_states(const key &id, std::vector<int> &groups)
...
 
 async_iterator_result session::start_iterator(const key &id, const std::vector<dnet_iterator_range>& ranges,
-								uint32_t type, uint64_t flags,
-								const dnet_time& time_begin, const dnet_time& time_end)
+		uint32_t type, uint64_t flags, const dnet_time& time_begin, const dnet_time& time_end)
 {
-	auto ranges_size = ranges.size() * sizeof(ranges.front());
+	size_t ranges_size = 0;
+	if (ranges.size() != 0)
+		ranges_size = ranges.size() * sizeof(ranges.front());
```
shaitan: What is the difference between
```c++
size_t ranges_size = ranges.size() * sizeof(ranges.front());
```
and
```c++
size_t ranges_size = 0;
	if (ranges.size() != 0)
		ranges_size = ranges.size() * sizeof(ranges.front());
```
?

bioothod: `ranges.front()` is undefined if vector has no elements

bindings/cpp/logger.cpp
bioothod: That's the purpose of this patch - to eliminate possible bug if `dnet_node_unset_trace_id()` has been called multiple times

bindings/cpp/logger.cpp
bioothod: I'd better to remove preallocated array, it is a little bit faster

bindings/cpp/logger.cpp
bioothod: Increasing scope here is a bug actually, although it never reached that level in our code.
Which I found by changing `DNET_SCOPED_LIMIT` and looking for crash in tests.
I'm against asserts (this is clearly a simplified version of assert) and this code can easily continue working without assertion, so I'd better revert this commit

shaitan:
Can you, please, split this PR? It is too big for reviewing, we could miss something important.

bindings/cpp/session.cpp
abudnik: sizeof(T) is compile-time operation and it doesn't depend on ranges.front() runtime value?

bindings/cpp/logger.cpp
shaitan: What changes do you suggest?

bioothod: I will return this chunk to how it was before

bindings/cpp/logger.cpp
shaitan: It doesn't eliminate the bug, it silently hides the bug.

bindings/cpp/session.cpp
bioothod: This is not about sizeof(), but about ranges.front(), which may fail.
Consider this as following example
```
int a[3] = {0, 1, 2};
printf("%d\n", sizeof(a[12]))'
```

Will it compile? Sure.
Will it fail? Probably not, since a[12] may or may not be dereferenced.

To make things clear I made those changes.

tests/server_send.cpp
```diff
@@ -0,0 +1,349 @@
...
+{
+	s.set_trace_id(rand());
+	for (int i = 0; i < num; ++i) {
+		std::string id = id_prefix + lexical_cast(i);
+
+		ELLIPTICS_REQUIRE_ERROR(res, s.read_data(id, 0, 0), error);
+	}
+}
+
+static std::vector<dnet_raw_id> ssend_ids(session &s)
```
abudnik: Can you, please, describe what should this function do?

bindings/cpp/logger.cpp
bioothod: You propose to leave it as is and possibly crash at random? No.
This is exactly the same check as in `std::vector::at()` method, it doesn't eliminate the bug, but it doesn't crash

tests/server_send.cpp
bioothod: It returns vector of unique IDs associated with all backends which live in groups assigned to given session, one per backend. Those ids can be used when one has to address all backends in specified groups, like my tests do.

bindings/cpp/logger.cpp
shaitan: `std::vector::at()` throws `out_of_range` if passed position is out of bounds.

bindings/cpp/session.cpp
abudnik: AFAIK, in case of sizeof(Expression), Expression is never evaluated.

tests/server_send.cpp
abudnik: Does it mean that backen_id for some group g1 is unique among all groups {Gi}? In other words, any two groups couldn't have two backends with the same backend_id value?

bindings/cpp/logger.cpp
bioothod: More important, it doesn't crash. Which does my check.

I had a version, which printed error message into log, but it is as useless as above errors in `dnet_node_set_trace_id()`, nobody sees it and it doesn't help finding the bug.

I do not understand, what is this about? You want me to revert this commit and allow possible crash? No, I will not do this, crash is not allowed compared to different attributes in the log. So, what is this about?

tests/server_send.cpp
bioothod: This is not about any two groups, it is about single source group where iteration runs.
And you likely mean not backend_id, but dnet_raw_id. Yes, ids are unique, that's achieved by route table, one has to select those ids which correspond to unique backends

bindings/cpp/session.cpp
bioothod: Is this part of standard? I do not know for sure, thus there is this change

tests/server_send.cpp
abudnik: Ok, I thought that `if (std::find(groups.begin(), groups.end(), entry.group_id) != groups.end()) {` implies multiple groups with all unique backen_id.

bindings/cpp/logger.cpp
shaitan: This is about hiding the bug without any words. It is bad practice.

bindings/cpp/session.cpp
abudnik: From C++14 standard:
5.3.3.1:
The sizeof operator yields the number of bytes in the object representation of its operand. The operand is either an expression, which is an unevaluated operand (Clause 5), or a parenthesized type-id.

bindings/cpp/logger.cpp
bioothod: What do you suggest besides telling that it is a bad practice?
Crash (as well as assert which leads to crash) is unacceptable.
Leaving potential bug as it was before is unacceptable either.

bindings/cpp/session.cpp
```diff
@@ -2249,7 +2266,42 @@ async_iterator_result session::start_iterator(const key &id, const std::vector<d
...
+	req->time_begin = time_begin;
+	req->time_end = time_end;
+	req->range_num = ranges.size();
+	req->group_num = dst_groups.size();
+
+	if (ranges_size)
+		memcpy(data.skip<dnet_iterator_request>().data(), &ranges.front(), ranges_size);
+
+	if (groups_size)
+		memcpy(data.skip(ranges_size + sizeof(dnet_iterator_request)).data(), &dst_groups.front(), groups_size);
```
shaitan: This partly duplicates `session::start_iterator`. Should `session::start_iterator` call `start_copy_iterator` with empty `dst_groups`?

bindings/cpp/session.cpp
bioothod: Ok, make sense. This can be returned to older code

bindings/cpp/logger.cpp
shaitan: I suggest to log a message on the ERROR level. It will help us in case of the bug.

bindings/cpp/session.cpp
bioothod: This will break API since `start_iterator` uses default parameters for time ranges.
I also added the same check as in plain server-send copy - iterator should return early if there are destination groups. This check is meaningless for other iterator types.

bindings/cpp/logger.cpp
bioothod: This forces changing all instances of `dnet_node_unset_trace_id()`, I had that commit, but removed, it is really useless.

I'm not totally against it, but let's separate it into its own pull request, this one unfortunately already contains enough unrelated changes.

bindings/cpp/session.cpp
shaitan: How calling `start_copy_iterator` inside of `start_iterator` will break API?

bioothod: `start_iterator` will have to accept new parameter, that's API change.
One can not call `start_copy_iterator` inside of `start_iterator` since the former checks destination groups array, it returns early if there is nothing to copy.

bindings/cpp/logger.cpp
shaitan: How about moving all refactoring and bug fixes into separate pull requests? It will help us to concentrate on new feature in this pull request's conversation.

bioothod: Unfortunately I can not easily separate refactoring and more notably bug fixes from new code, since new code bugfixes also fixed possible old issues

bindings/cpp/logger.cpp
```diff
@@ -137,13 +137,12 @@ void dnet_node_set_trace_id(dnet_logger *logger, uint64_t trace_id, int tracebit
 	if (scoped_count >= DNET_SCOPED_LIMIT) {
 		dnet_log_only_log(logger, DNET_LOG_ERROR,
 			"logic error: you must not call dnet_node_set_trace_id twice, dnet_node_unset_trace_id call missed");
-		++scoped_count;
```
shaitan: All calls of `dnet_node_set_trace_id()` are followed by calls of `dnet_node_unset_trace_id()`. So if it doesn't increment `scoped_count` here the following call of `dnet_node_unset_trace_id()` will unset someone's, but not own, trace_id.

bindings/cpp/session.cpp
shaitan: No, `start_iterator` will not. Nothing with API should be changed. Only `start_iterator`'s implementation should be changed.

bindings/cpp/logger.cpp
bioothod: Ok, makes sense

bindings/cpp/session.cpp
bioothod: Then you can not return if destination group array is empty.
Or you will have to check iterator type and then check group array if it is copy iterator, which breaks decomposition.
I'd rather remove type argument from `start_copy_iterator`, it is useless

bindings/cpp/logger.cpp
shaitan: But the alternative - this PR will have batch of incomplete unrelated changes which will be complemented in separate pull requests.

bindings/cpp/session.cpp
shaitan: I can't see any checks within `start_copy_iterator` which makes it different from `start_iterator`.

bindings/cpp/logger.cpp
bioothod: This pull request fixes set of possible bugs.
It also fixes real bugs.

In an ideal world those changes should be separated, but in real life we deal with our best to separate things and are limited by that.

I wish everyone spent their time fixing bugs with such scrutiny which attracted these innocent changes.

bindings/cpp/session.cpp
bioothod: `start_iterator` will always have empty destination groups, if you call `start_copy_iterator` from `start_iterator`, it will always return without running non-copy iterator.

shaitan: The only differences between `start_iterator` and `start_copy_iterator`:
`start_iterator` doesn't set `req->group_num` (leaves it zero) and doesn't append groups into `data_pointer data`.

There is no more different, so if I call `start_copy_iterator` with empty groups it will be performed as `start_iterator` with corresponding arguments.

bindings/cpp/logger.cpp
```diff
@@ -156,32 +156,37 @@ void dnet_node_set_trace_id(dnet_logger *logger, uint64_t trace_id, int tracebit
...
 		}
 
+
 		new (local_attributes) scoped_attributes_t(*logger, std::move(attributes));
 
 		// Set all bits to ensure that it has tracebit set
 		backend_trace_id_hook = scoped_trace_id_hook[scoped_count];
-	} catch (...) {
-		local_attributes = NULL;
+		++scoped_count;
```
shaitan: It is the same case, `scoped_count` have to be incremented even in case of exception.

bioothod: Ok, makes sense

bindings/cpp/session.cpp
bioothod: The latter is invalid, copy iterator has to copy data, that is its mission.
That's why it checks whether destination group array is empty and returns early.

I've removed type from `start_copy_iterator` method, it is always the same.

shaitan: `start_copy_iterator` and `start_iterator` have a lot in common. Could this common part be moved into helper function? Because if we will found a bug in this common code we will have to double check and double fix it. For example,  `start_copy_iterator` doesn't memset data like `start_iterator` do. With helper function you will avoid such problem.

include/elliptics/session.hpp
```diff
@@ -671,10 +679,18 @@ class session
...
+		async_iterator_result start_copy_iterator(const key &id, const std::vector<dnet_iterator_range>& ranges,
+								uint64_t flags,
+								const dnet_time& time_begin,
+								const dnet_time& time_end,
+								const std::vector<int> &dst_groups);
 		async_iterator_result pause_iterator(const key &id, uint64_t iterator_id);
 		async_iterator_result continue_iterator(const key &id, uint64_t iterator_id);
 		async_iterator_result cancel_iterator(const key &id, uint64_t iterator_id);
 
+		async_iterator_result server_send(const std::vector<std::string> &keys, uint64_t flags,
```
shaitan: `start_copy_iterator` and `server_send` from public API have to be documented

bindings/cpp/session.cpp
bioothod: All common part already lives in `iterator()`, it has to be initialized, and in this routing there are some differences.
They are quite minor, but they are real.

Helper function will accept 10 parameters and then set the control structure.
Instead we set control structures in the public method and just pass it to iterator.

include/elliptics/session.hpp
bioothod: Btw, you forgot to document other iterator functions, I will write doc comment for all

bindings/cpp/session.cpp
shaitan: The control structure is ok, but it should be refactored in the way that allows to escape repeating yourself. Currently it doesn't and makes a good ground for bugs.

tests/server_send.cpp
```diff
@@ -0,0 +1,349 @@
...
+		copied++;
+	}
+
+	BH_LOG(log, DNET_LOG_NOTICE, "%s: keys: %d, dst_groups: %s, copied total: %d",
+			__func__, num, print_groups(dst_groups), copied);
+
+	BOOST_REQUIRE_EQUAL(copied, num);
+}
+
+static bool ssend_register_tests(test_suite *suite, node &n)
```
abudnik: What if flag DNET_IFLAGS_MOVE will not be set. Should it copy or move data?

bioothod: It will copy data, this is described in flags description

abudnik: Can you please add additional test to check this logic?

bindings/cpp/session.cpp
bioothod: It is already a control structure setup.
I do not understand which bugs you are talking about in the code which fills in members of the structure from function arguments. This can not be made more simpler than that.

tests/server_send.cpp
bioothod: Which logic? Everything is already being tested.

abudnik: If flag DNET_IFLAGS_MOVE is not set, then all keys must be in src group after copying them to dst group.

bioothod: You read the code wrong, it is already being tested

bindings/cpp/session.cpp
shaitan: For example, `start_copy_iterator` doesn't memset `data` by 0, so it leaves some part of control structure uninitialized.

bindings/cpp/session.cpp
```diff
@@ -2287,6 +2340,110 @@ async_iterator_result session::cancel_iterator(const key &id, uint64_t iterator_
...
+	}
+
+	if (keys.empty()) {
+		async_iterator_result result(*this);
+		async_result_handler<iterator_result_entry> handler(result);
+		handler.complete(create_error(-ENXIO, "server_send: key list is empty"));
+		return result;
+	}
+
+	int local_group = get_groups().front();
```
shaitan: I know that we have the same in all iterator methods, but we have to add to iterator methods' and server_send's descriptions that they use only first group from groups specified in the session.

tests/server_send.cpp
abudnik: It seems to me that flag DNET_IFLAGS_MOVE is _always_ set in ssend_register_tests(). Am I right?

bindings/cpp/session.cpp
```diff
@@ -2287,6 +2340,110 @@ async_iterator_result session::cancel_iterator(const key &id, uint64_t iterator_
...
+			if (cmp < 0)
+				return true;
+			if (cmp > 0)
+				return false;
+			return backend_id < other.backend_id;
+		}
+	};
+
+	std::map<la, std::vector<dnet_raw_id>> raw_ids;
+	for (const auto &key : keys) {
```
shaitan: Do we support lucid? This won't be compiled on lucid.

tests/server_send.cpp
bioothod: Because this is additional to copy logic. It is copy + remove. Which means copy is already tested by reading data from remote nodes.

This becomes useless, is this anything you can comment on 2k-lines patch?

bindings/cpp/session.cpp
shaitan: gcc 4.4.7 from RHEL6 doesn't support this too(

bindings/cpp/session.cpp
bioothod: It doesn't fill it with zeroes, because it fills all members already.
But I added request structure memset.

This arguing around API doesn't move forward. I will not implement additional helper which will accept 10 arguments and setup control structure. Instead I implement 2 different methods which setup control structures. And that's how things have to be done, unfortunately we already have too many functions with 10+ arguments.

Let's stop talking about API and try to find out possible bugs in other places.
If it is the only thing you care about, then this review is useless.

bindings/cpp/session.cpp
```diff
@@ -2287,6 +2340,110 @@ async_iterator_result session::cancel_iterator(const key &id, uint64_t iterator_
...
+		la l;
+
+		transform(key, l.id);
+
+		err = dnet_lookup_addr(get_native(), NULL, 0, &l.id, local_group, &l.addr, &l.backend_id);
+		if (err != 0) {
+			async_iterator_result result(*this);
+			async_result_handler<iterator_result_entry> handler(result);
+			handler.complete(create_error(-ENXIO,
```
shaitan: Why doesn't it use err to notify user about the original problem?

bindings/cpp/session.cpp
bioothod: Ok, I've replaced it with usual iterator

tests/server_send.cpp
abudnik: If I wrote this tests, I would add some test like:
1. ssend_test_insert_many_keys(src, keys)
2. ssend_test_copy(src, dest, keys, flags) // flags without DNET_IFLAGS_MOVE
3. ssend_test_read_many_keys(src) // all keys must be there

bindings/cpp/session.cpp
bioothod: It is ENXIO already, but if it would be different, iterateor's err has to be consistent with 'no group' route table error, which is a common -6 error in our code

tests/server_send.cpp
bioothod: Please, read the tests below.

bioothod: I'm curios do you really check that after you've read the key you check that key is still there?
Please stop this stupidity.

bindings/cpp/session.cpp
bioothod: Ok, I'll update comments

bindings/cpp/session.cpp
shaitan: Please, be patient and more constructive. You have made a great job and now it should be carefully reviewed.

bindings/cpp/session.cpp
shaitan: If there is no difference, why don't use original error code? If someone change `dnet_lookup_addr` implementation it will not affect to this error.

bindings/cpp/session.cpp
```diff
@@ -2287,6 +2341,110 @@ async_iterator_result session::cancel_iterator(const key &id, uint64_t iterator_
...
+		auto it = raw_ids.find(l);
+		if (it == raw_ids.end()) {
+			l.id.group_id = local_group;
+			raw_ids[l] = std::vector<dnet_raw_id>({raw});
+		} else {
+			it->second.push_back(raw);
+		}
+	}
+
+	const size_t groups_size = groups.size() * sizeof(int);
```
shaitan: It is better to use `sizeof(groups.front())` because it avoids changing groups type without changing type in this sizeof.

bindings/cpp/session.cpp
```diff
@@ -2287,6 +2341,110 @@ async_iterator_result session::cancel_iterator(const key &id, uint64_t iterator_
...
+
+	std::list<async_iterator_result> results;
+	for (const auto &it: raw_ids) {
+		auto &id = it.first;
+		auto &ids = it.second;
+
+		const size_t ids_size = ids.size() * sizeof(dnet_raw_id);
+
+		data_pointer data = data_pointer::allocate(sizeof(dnet_server_send_request) + ids_size + groups_size);
```
shaitan: Just a hint: you can use `data_buffer` instead of `data_pointer` here like it is done in [session_indexes.cpp#L111](https://github.com/reverbrain/elliptics/blob/master/bindings/cpp/session_indexes.cpp#L111). It allows to simplify code and to avoid offset calculation with `memcpy`.

bindings/cpp/session.cpp
bioothod: As you said, someone will change `dnet_lookup_addr()` and there will no longer be `-6` error when there is no appropriate entry in the route table. And we are used to this this error, it is a defacto code for this kind of problems.

shaitan: Makes sense

bindings/cpp/session.cpp
bioothod: `data_buffer` doesn't have offsets, it should be either overwritten from the beginning by contiguous chunks, or introduce additional pointer and dereference it.

library/dnet.c
```diff
@@ -393,6 +406,375 @@ static int dnet_iterator_callback_send(void *priv, void *data, uint64_t dsize)
...
+{
+	if (atomic_dec_and_test(&ctl->refcnt))
+		return dnet_server_send_cleanup(ctl);
+
+	return 0;
+}
+
+static int dnet_server_send_sync(struct dnet_server_send_ctl *ctl)
+{
+	while (atomic_read(&ctl->writes_pending) > 0) {
```
abudnik: I hope that this loop will not be executed for a long time, because __need_exit flag may be set after program interruption.

bioothod: When someone sets `st->__need_exit` flag appropriate network or IO thread starts killing the state and removes all pending transactions, which in turn invoke server-send completion callbacks which decrease this variable.

There is intermediate sync which waits until 500 writes out of 1000 pending operations complete, it checks `st->__need_exit` flag though, since it can be interrupted in the middle of the iteration, while `dnet_server_send_sync()` is only called to wait for the last 1000 writes to complete.

bindings/cpp/session.cpp
shaitan: This code fills `data` from the beginning by contiguous chunks, so it can be easily rewritten with use of `data_buffer`. But, as I said, it is just a hint.

tests/server_send.cpp
shaitan: You've made a copy operation which with certain flag removes data on the source. Where is the test or the guarantee that without this certain flag this copy operation will not remove data on the source?

Please, be more constructive. All sentences with 'I will not', 'stop this stupidity' etc. will be ignored, because they doesn't make sense.

example/eblob_backend.c
```diff
@@ -872,13 +874,156 @@ int blob_defrag_start(void *priv, enum dnet_backend_defrag_level level)
...
+
+	/* structure has been already checked and has been proved to be correct no need to perform sanity checks */
+	dnet_convert_server_send_request(req);
+
+	ids = (struct dnet_raw_id *)(req + 1);
+	groups = (int *)(ids + req->id_num);
+
+	memset(&re, 0, sizeof(struct dnet_iterator_response));
+	re.total_keys = req->id_num;
+	re.flags = req->iflags;
```
shaitan: Why do you set `re.flags` to `req->iflags`? `re.flags` is designed to be set to iterated key flags from the backend.

bindings/cpp/session.cpp
```diff
@@ -2287,6 +2341,110 @@ async_iterator_result session::cancel_iterator(const key &id, uint64_t iterator_
...
+async_iterator_result session::server_send(const std::vector<std::string> &keys, uint64_t iflags, const std::vector<int> &groups)
+{
+	if (get_groups().empty()) {
+		async_iterator_result result(*this);
+		async_result_handler<iterator_result_entry> handler(result);
+		handler.complete(create_error(-ENXIO, "server_send: local group list is empty"));
+		return result;
+	}
+
+	if (groups.empty()) {
```
shaitan: What will happen if `groups` contain `local_group`?

tests/server_send.cpp
bioothod: You do not really help to make this code better with this. You stop progress.
I wish we spent this time searching for bugs. I'm curious how many tests you will write for the next PR.

If you do not understand, I will raise different question - where it test or guarantee that plain iterator doesn't remove data? Why didn't you implement this test for iterator?

bindings/cpp/session.cpp
bioothod: Sigh. It doesn't.
It fills the beginning of the data with response structure members one-by-one. Or memsetting it.
And writes data with offset afterwards. This operation is not supported. One has to setup separate response structure and then copy it first, then ids and then data.

This is unneeded copy.

example/eblob_backend.c
bioothod: Updated.

bindings/cpp/session.cpp
bioothod: It will write data to itself

tests/server_send.cpp
shaitan: >You do not really help to make this code better with this. You stop progress.
I wish we spent this time searching for bugs.

Really? We've already pointed you on potential bugs and that your tests don't cover cases which lead to data lost. Review is the time to stop, understand and analyze code of new feature. And you don't help us with your ironic and reluctance to split this 2k+-lines patch where mixed new feature and unrelated refactoring and bugs fixes.

> I'm curious how many tests you will write for the next PR.
Let's not measure the number of tests, but the quality. And again, please, avoid using such intonation in your words, it doesn't make anything.

>where it test or guarantee that plain iterator doesn't remove data? Why didn't you implement this test for iterator?
Plain iterator doesn't remove data in any case with any flags and any types, there is no code which could remove data in plain iterator. Copy iterator with certain flag removes data, so it contains code which removes data and it should be tested that copy iterator doesn't remove data without this certain flag.

bindings/cpp/session.cpp
shaitan: You should try to compile this PR on lucid or RHEL6 because it contains a lot of code that won't be compiled.

tests/server_send.cpp
bioothod: What do you mean by `potential bug`? Is it data corruption, or maybe crash, which have been found here?

This PR fixes couple of such things, and I didn't separate them because it is hard, and efforts on doing this do not worth amount of time needed. Apparently you do not count utility of the time discussing this things.

Let me explain the nature of the test. Test tests things which code touches. If code removes data, one has to test whether data has been removed. If code doesn't remove data (like plain iterators and copy iterator), none tests this feature. Move iterator removes data, so there is a test, which checks exactly that - whether data has been removed or not. When copy iterator runs with certain set of flags, it removes data and this is tested. When it runs with different set of flags, it doesn't touch data, testing that it doesn't touch data is useless.

I'm tired wasting time with it. I've added this test. I do expect the same level of scrutiny for any other code now, I hope it will help to find bugs before next such hackathon.

bindings/cpp/session.cpp
shaitan: Is it expected?

In some case it could be useful, for example, if you want to convert data locally, but it should be additionally described in the docs with details and examples. It isn't stop requirement for this pull request, but it is recommendation from the category "good to have".

bioothod: Why not? I didn't think about this case, but why not?

bindings/cpp/session.cpp
bioothod: Updated

tests/server_send.cpp
shaitan: > What do you mean by potential bug? Is it data corruption, or maybe crash, which have been found here?

`Potential bug` - bug that was fixed within this review.

>This PR fixes couple of such things, and I didn't separate them because it is hard, and efforts on doing this do not worth amount of time needed. Apparently you do not count utility of the time discussing this things.

You do not count utility of our time when we try to keep this unrelated changes in mind and tries to analyze them separately.

> I'm tired wasting time with it. I've added this test. I do expect the same level of scrutiny for any other code now, I hope it will help to find bugs before next such hackathon.

Thanks. We always have the same level of scrutiny for all changes which could be merged into master.

bindings/cpp/session.cpp
shaitan: So it is the feature, not a bug. That's it.

library/dnet.c
abudnik: Also, it is possible to rewrite this loop to smth like this (to minimize syscalls):
1. mutex lock
2. while(pending>0) { cond_wait(cond, mutex) }
3. mutex unlock

library/dnet.c
```diff
@@ -393,6 +406,375 @@ static int dnet_iterator_callback_send(void *priv, void *data, uint64_t dsize)
...
+	ctl.io.timestamp = re->timestamp;
+	ctl.io.user_flags = re->user_flags;
+	ctl.io.total_size = re->size;
+	ctl.io.size = re->size;
+	ctl.io.flags = DNET_IO_FLAGS_WRITE_NO_FILE_INFO;
+
+	// overwrite doesn't care whether remote key differs from local
+	// when this flag is not set, we only overwrite the same data or if there is no remote copy at all
+	if (!(send->iflags & DNET_IFLAGS_OVERWRITE))
+		ctl.io.flags |= DNET_IO_FLAGS_COMPARE_AND_SWAP;
```
shaitan: `DNET_IO_FLAGS_COMPARE_AND_SWAP` makes node to compare checksum of stored data with checksum from dnet_io_attr.parent. Node will make the operation only if it doesn't have the key or checksums are equal.

As I see, you have unfilled `ctl.io.parent`, so this write operation will succeed only if the key is missed on the remote node. So without `DNET_IFLAGS_OVERWRITE` it doesn't overwrite keys at all, it writes only new keys.

library/dnet.c
```diff
@@ -393,6 +406,375 @@ static int dnet_iterator_callback_send(void *priv, void *data, uint64_t dsize)
...
+	// deliberately do not set DNET_FLAGS_NEED_ACK
+	// if WRITE command has failed, @dnet_process_cmd_with_backend_raw() will set this bit automatically,
+	// and will send acknowledge with error
+	//
+	// if there is no error, @dnet_file_info structure will be returned
+	ctl.cflags = 0;
+
+	ctl.fd = fd;
+	ctl.local_offset = data_offset;
+	ctl.cmd = DNET_CMD_WRITE;
```
shaitan: Sending full data of all keys to all destination groups which will heavy utilize the network for sending data of keys whose write command will be failed on checksums checking.

library/dnet.c
```diff
@@ -393,6 +406,375 @@ static int dnet_iterator_callback_send(void *priv, void *data, uint64_t dsize)
...
+		ctl.io.flags |= DNET_IO_FLAGS_COMPARE_AND_SWAP;
+
+	// deliberately do not set DNET_FLAGS_NEED_ACK
+	// if WRITE command has failed, @dnet_process_cmd_with_backend_raw() will set this bit automatically,
+	// and will send acknowledge with error
+	//
+	// if there is no error, @dnet_file_info structure will be returned
+	ctl.cflags = 0;
+
+	ctl.fd = fd;
```
shaitan: Using `fd` here is ok, but remote node will keep whole data in-memory that may cause the node will be killed by oom killer.

abudnik:
**Merge dc tool optimization**

I think that using server send operations could optimize merge dc tool: you can iterate every node, merge all keys and find latest keys among them (nothing has changed here since current merge dc implementation). If some key at node1 has latest timestamp and other replicas have older timestamp at node2, node3, then you can call server send operation for this key at node1 with {node2, node3} destination groups as a parameter. This method may be optimized for calling server send only once per replica, but main idea is elimination of intermediary node (where merge dc called) for migrating data.

shaitan:
Let's do not involve recovery into this pull request, it already has enough unrelated things.

library/dnet.c
```diff
@@ -314,6 +314,31 @@ int dnet_send_reply(void *state, struct dnet_cmd *cmd, const void *odata, unsign
 	return err;
 }
 
+void dnet_queue_wait_threshold(void *state)
+{
+	struct dnet_net_state *st = state;
```
shaitan: What is the reason for passing `state` as `void *`?

library/dnet.c
```diff
@@ -393,6 +406,375 @@ static int dnet_iterator_callback_send(void *priv, void *data, uint64_t dsize)
...
+					__func__,
+					dnet_dump_id(&send->cmd.id), dnet_dump_id_str(re->key.id),
+					(unsigned long long)re->user_flags,
+					(unsigned long long)re->timestamp.tsec, (unsigned long long)re->timestamp.tnsec,
+					re->status, (unsigned long long)re->size,
+					(unsigned long long)re->iterated_keys, (unsigned long long)re->total_keys,
+					send->write_error);
+
+			atomic_dec(&send->writes_pending);
+			pthread_cond_broadcast(&send->write_wait);
```
shaitan: Broadcasting conditional variable after each transaction completion will cause to false-positive awakening of the thread which is waiting until `writes_pending` is less then 500.

include/elliptics/packet.h
```diff
@@ -1145,6 +1218,21 @@ static inline void dnet_convert_monitor_stat_request(struct dnet_monitor_stat_re
 	r->categories = dnet_bswap32(r->categories);
 }
 
+
+struct dnet_server_send_request {
+	int		id_num;
+	int		group_num;
+	uint64_t	iflags;
+};
```
shaitan: Shouldn't `dnet_server_send_request` have `__attribute__ ((packed))`?

library/dnet.c
```diff
@@ -1013,6 +1484,9 @@ static int dnet_process_cmd_with_backend_raw(struct dnet_backend_io *backend, st
 	gettimeofday(&end, NULL);
 	diff = DIFF(start, end);
 
+	/* If there was any error - send ACK to notify client with error code and destroy transaction */
+	if (err)
+		cmd->flags |= DNET_FLAGS_NEED_ACK;
```
shaitan: Have you checked that all other commands don't send ACK with error by themselves?

library/dnet.c
```diff
@@ -393,6 +406,375 @@ static int dnet_iterator_callback_send(void *priv, void *data, uint64_t dsize)
...
+{
+	int err;
+	struct dnet_net_state *st = state;
+	struct dnet_server_send_ctl *ctl;
+
+	dnet_state_get(st);
+
+	ctl = malloc(sizeof(struct dnet_server_send_ctl) + sizeof(int) * group_num);
+	if (!ctl) {
+		err = -ENOMEM;
```
shaitan: No one uses this err, so this can be omitted.

library/dnet.c
bioothod: Well, yes, what is the purpose of overwriting keys with the same data?
Maybe only timestamp update, but I doubt anyone uses that

library/dnet.c
bioothod: No node will not keep data in memory, what are you talking about?

library/dnet.c
bioothod: That's is the purpose of compare-and-swap write - fail it if checksum mismatch.
But if it matches, what is the purpose of overwriting data in the first place?

library/dnet.c
shaitan: It doesn't check that another side has the same data or not.

library/dnet.c
bioothod: Maybe they do, that would be a bug.

This particular patch fixes timeouts for write errors with compare-and-swap and failed checksums, since write never sends any reply in this case. This is also true for all other commands which do not explicitly set NEED_ACK bit.

library/dnet.c
bioothod: `struct dnet_net_state` is private internal structure, it can not be operated on by backends, which pass it as a void pointer.
It is the same logic as with `dnet_send_reply()` and friends

library/dnet.c
shaitan: Remote node will read whole packet from socket into memory which includes reading whole key's data.

library/dnet.c
shaitan: Failed requests will utilize network for nothing.

include/elliptics/packet.h
bioothod: No, it should not, since it is already evenly packed to the highest member size.

Packing is a very bad idea generally, unaligned access is very expensive on the most but x86 architectures, it should only be used when we absolutely have to because of badly designed structures.

library/dnet.c
bioothod: This is a very bad idea to remove error assignment just because it is not used.
Eventually we may want to print it or return and may forget that it was not assigned in some error path.

library/dnet.c
bioothod: That's why there is overwrite flag. It tells to overwrite data no matter what.
When there is no overwrite flag, it should only write data if it doesn't exist or if it is the same, but what is the purpose of overwriting the same data?

Saying with other words, if one puts a checksum, it will write either new data to remote node, or overwrite the same. But there is no sense in overwriting the same data.

library/dnet.c
shaitan: By default, all handlers send ACK with error by themselves. READ, WRITE and LOOKUP were exceptions because they could be handled by differently (by cache or by backend) - it was solved via expecting that high level function (`dnet_process_cmd_with_backend_raw`) will send ACK for them.

With these changes, all commands shouldn't send ACK by themselves, so it should be carefully rechecked.

library/dnet.c
bioothod: That's how write command works, yes. Or do you propose splitting data into chunks and sending them with multiple write commands using prepare/commit logic? Make it more clear.

That could be possible, although I'm not sure that it is so much needed.

library/dnet.c
bioothod: Again, that is the purpose of compare and swap write - fail it if checksum mismatch.
Yes, it will utilize the bandwidth. Yes, it will require reading data from disk.

The main purpose of iterator sending all of its data or some ranges of its data to remote nodes is to do this during recovery, in particular case when you do know that remote group is empty after crash or you are sure it doesn't contain content from this *local* node. It is a replacement for `rsync` and optionally for merge recovery, but in this case its main purpose is not only to send data but also to remove it locally.

If remote group is not empty, sending bulk load of writes will not be effective since most of the keys already exist.
That is why there is `session::server_send()` method which operates on per-key basis.

library/dnet.c
bioothod: No command sends ack itself ever (btw, sending ack means sending small packet without data and without DNET_FLAGS_MORE bit), it is always being sent at the very end of `dnet_process_cmd_raw()` for every commands.
IO commands do not send ack at all, since they send data, and that's already a sign, that everything is ok.
They have to send ack when there are errors though. There must be an ack if any command has failed, since in this case command itself doesn't send data.

library/dnet.c
shaitan: Forget about overwrite flag. With default flags it will write keys only if they aren't exist on remote node. With default flags it will not overwrite anything.

bioothod: Exactly! That's the goal.

library/dnet.c
bioothod: Original sender is also waiting for state changes too.
It is not a performance penalty though compared to amount of work needed to send data over the network, but it can be cleaned up.

library/dnet.c
shaitan: Is it really needed to call `dnet_queue_wait_threshold` by backend? It is called only by `dnet_send_reply_threshold`, so it can be defined as a static.

library/dnet.c
bioothod: Mutex is not a syscall until it has to go into the kernel and sleep, in which case there is no much gain in optimizing that sleep. But this check can still be optimized, I will update patch

library/dnet.c
shaitan: You should change description for pull request and comments for this feature, because it says: "writing data to remote node using compare-and-swap, i.e. only write data if it either doesn't exist or it is the same on remote servers".

library/dnet.c
shaitan: `session::server_send()` isn't suitable because it accepts only original (std::string) keys.

library/dnet.c
bioothod: I used this function in implementing server-send command in eblob originally. It is not needed now and can be made static

library/dnet.c
shaitan: Yes, it is better to split data into chunks, because this will fail on big keys.

library/dnet.c
bioothod: That is a reason to extend API and not nitpicking

library/dnet.c
bioothod: Updated

library/net.c
```diff
@@ -599,29 +599,32 @@ static int dnet_process_update_ids(struct dnet_net_state *st, struct dnet_cmd *c
...
-		err = -ENOMEM;
-		goto err_out_exit;
-	}
-
-	err = dnet_validate_id_container(container, cmd->size, backends);
+	err = dnet_validate_id_container(id_container, cmd->size);
 	if (err) {
 		dnet_log(st->n, DNET_LOG_ERROR, "failed to validate route-list container from state: %s, err: %d",
 			dnet_state_dump_addr(st), err);
 		goto err_out_free;
```
shaitan: This will lead to free of uninitialized `backends`.

library/dnet.c
```diff
@@ -669,6 +1057,38 @@ static int dnet_iterator_start(struct dnet_backend_io *backend, struct dnet_net_
...
+		 *
+		 * When we clear NEED_ACK bit here, we prevent sending ACK
+		 * when refcnt reaches zero. We have to restore bit to allow
+		 * command completion to send ACK.
+		 *
+		 * It is safe to change bit, since @dnet_server_send_alloc() copies
+		 * dnet_cmd, thus it will store command structure without NEED_ACK bit.
+		 */
+		cmd->flags &= ~DNET_FLAGS_NEED_ACK;
+		sspriv = dnet_server_send_alloc(st, cmd, ireq->flags, dst_groups, ireq->group_num);
```
shaitan: There's no check that `sspriv` isn't `NULL`.

library/elliptics.h
```diff
@@ -1027,6 +1027,40 @@ static inline int dnet_read_ll(int fd, char *data, size_t size, off_t offset)
...
+	uint64_t			iflags;		/* Iterator flags */
+
+	int				*groups;	/* Groups to send WRITE commands */
+	int				group_num;
+
+	pthread_mutex_t			write_lock;	/* Lock for @write_wait */
+	pthread_cond_t			write_wait;	/* Waiting for pending writes */
+	atomic_t			bytes_pending;	/* Number of bytes in-flight to remote servers */
+
+	int				write_error;	/* Set to the first error occured during write
```
shaitan: "Set to the first error **occurred** during write"

library/dnet.c
```diff
@@ -393,6 +404,375 @@ static int dnet_iterator_callback_send(void *priv, void *data, uint64_t dsize)
...
+
+	s = dnet_session_create(n);
+	if (!s) {
+		dnet_log(n, DNET_LOG_ERROR, "%s: Interrupting iterator because failed to create new session",
+				dnet_dump_id(&send->cmd.id));
+		err = -ENOMEM;
+		goto err_out_free;
+	}
+
+	dnet_session_set_trace_id(s, send->cmd.trace_id);
```
shaitan: Should it also copy send->cmd.flags? At least it should copy trace_bit flag and may be some other.

example/eblob_backend.c
```diff
@@ -872,13 +874,156 @@ int blob_defrag_start(void *priv, enum dnet_backend_defrag_level level)
...
+
+		err = blob_lookup(b, &key, &wc);
+		if (err < 0) {
+			dnet_backend_log(cfg->blog, DNET_LOG_ERROR, "%s: EBLOB: blob_send: lookup: %d: %s",
+					 dnet_dump_id_str(key.id), err, strerror(-err));
+			goto err_out_send_fail_reply;
+		}
+
+		re.key = ids[i];
+		re.flags = wc.flags; // these flags corresponse to DNET_RECORD_FLAGS_*
```
shaitan: "these flags **correspond** to DNET_RECORD_FLAGS_*"

library/dnet.c
```diff
@@ -393,6 +404,375 @@ static int dnet_iterator_callback_send(void *priv, void *data, uint64_t dsize)
...
+
+	err = dnet_session_set_groups(s, send->groups, send->group_num);
+	if (err) {
+		dnet_log(n, DNET_LOG_ERROR, "%s: Interrupting iterator because failed to set %d groups",
+				dnet_dump_id(&send->cmd.id), send->group_num);
+		err = -ENOMEM;
+		goto err_out_session_destroy;
+	}
+
+	atomic_add(&send->bytes_pending, re->size);
```
shaitan: It doesn't count that key is being written to one or to several groups.

library/dnet.c
```diff
@@ -393,6 +404,375 @@ static int dnet_iterator_callback_send(void *priv, void *data, uint64_t dsize)
...
+
+	dnet_log(n, DNET_LOG_NOTICE, "%s: %s: sending WRITE request, iterator response: %s, user_flags: %llx, ts: %lld.%09lld, "
+			"status: %d, size: %lld, iterated_keys: %lld/%lld",
+			__func__,
+			dnet_dump_id(&send->cmd.id), dnet_dump_id_str(re->key.id),
+			(unsigned long long)re->user_flags,
+			(unsigned long long)re->timestamp.tsec, (unsigned long long)re->timestamp.tnsec,
+			re->status, (unsigned long long)re->size,
+			(unsigned long long)re->iterated_keys, (unsigned long long)re->total_keys);
+	dnet_convert_iterator_response(re);
```
shaitan: Why is it needed here?

library/dnet.c
```diff
@@ -393,6 +404,375 @@ static int dnet_iterator_callback_send(void *priv, void *data, uint64_t dsize)
...
+	pthread_mutex_unlock(&ctl->write_lock);
+
+	return dnet_server_send_put(ctl);
+}
+
+/*
+ * Helper function which sends given data as WRITE command to remote groups
+ */
+int dnet_server_send_write(struct dnet_server_send_ctl *send,
+		struct dnet_iterator_response *re, uint64_t dsize,
```
shaitan: Can `dsize` be different from `sizeof(struct dnet_iterator_response)`?

example/eblob_backend.c
```diff
@@ -872,13 +874,156 @@ int blob_defrag_start(void *priv, enum dnet_backend_defrag_level level)
...
+		re.status = 0;
+		re.iterated_keys = i;
+		re.size = wc.total_data_size;
+		// set iterator response id to differentiate various commands
+		// client can use cmd->backend_id from reply though
+		re.id = cmd->backend_id;
```
shaitan: If it uses common iterator responses, can it be paused, continued and/or canceled via iterator's API?

example/eblob_backend.c
```diff
@@ -872,13 +874,156 @@ int blob_defrag_start(void *priv, enum dnet_backend_defrag_level level)
...
+
+		err = dnet_server_send_write(ctl, &re, sizeof(struct dnet_iterator_response), wc.data_fd, data_offset);
+		if (err)
+			goto err_out_send_fail_reply;
+
+		continue;
+
+err_out_send_fail_reply:
+		re.status = err;
+		err = blob_send_reply(state, cmd, &re, 1);
```
shaitan: Will it really skip all keys after it faces the first one which:
* doesn't exist
* or has broken ext header
* or has broken checksum
* or it has `dnet_server_send_write` failed

example/eblob_backend.c
```diff
@@ -872,13 +874,156 @@ int blob_defrag_start(void *priv, enum dnet_backend_defrag_level level)
...
+	ids = (struct dnet_raw_id *)(req + 1);
+	groups = (int *)(ids + req->id_num);
+
+	memset(&re, 0, sizeof(struct dnet_iterator_response));
+	re.total_keys = req->id_num;
+
+	/*
+	 * Set NEED_ACK bit to signal server-send controller that we want
+	 * to send final ACK when controller will be destroyed, which in turn
+	 * will happend after all WRITE commands are completed.
```
shaitan: "will **happened** after all WRITE commands are completed."

library/dnet.c
```diff
@@ -393,6 +404,375 @@ static int dnet_iterator_callback_send(void *priv, void *data, uint64_t dsize)
...
+					memset(r, 0, cmd_size);
+
+					r->header = r + 1;
+					r->hsize = sizeof(struct dnet_cmd);
+					r->data = r->header + sizeof(struct dnet_cmd);
+					r->dsize = sizeof(struct dnet_io_attr);
+					r->st = dnet_state_get(st);
+
+					lc = r->header;
+					dnet_setup_id(&lc->id, send->cmd.id.group_id, cmd->id.id);
```
shaitan: Previously, it checks `if (cmd)`, so I suppose that `cmd` can be `NULL`. This code will fail if `cmd` is `NULL`.

library/dnet.c
```diff
@@ -393,6 +404,375 @@ static int dnet_iterator_callback_send(void *priv, void *data, uint64_t dsize)
...
+					size_t cmd_size = sizeof(struct dnet_io_req) +
+							sizeof(struct dnet_cmd) +
+							sizeof(struct dnet_io_attr);
+
+					r = malloc(cmd_size);
+					if (!r) {
+						err = -ENOMEM;
+						// we could update wp->data here, which is dnet_iterator_response
+						// but we do not really care about local errors, for example
+						// remove error is not handled too
```
shaitan: Why do we ignore this local error?

example/eblob_backend.c
shaitan: Is it ok that it ignores `err`?

bindings/cpp/session.cpp
```diff
@@ -2249,7 +2264,46 @@ async_iterator_result session::start_iterator(const key &id, const std::vector<d
...
-	memcpy(data.skip<dnet_iterator_request>().data(), &ranges.front(), ranges_size);
+	if (ranges_size)
+		memcpy(data.skip<dnet_iterator_request>().data(), &ranges.front(), ranges_size);
+
+	return iterator(id, data);
+}
+
+async_iterator_result session::start_copy_iterator(const key &id,
+		const std::vector<dnet_iterator_range>& ranges,
+		uint64_t flags,
```
shaitan: Not all of `DNET_IFLAGS_*` may be used here. It should check that client pass only correct set of flags or unset inappropriate flags.

bindings/cpp/session.cpp
```diff
@@ -2233,12 +2248,12 @@ error_info session::mix_states(const key &id, std::vector<int> &groups)
 }
 
 async_iterator_result session::start_iterator(const key &id, const std::vector<dnet_iterator_range>& ranges,
-								uint32_t type, uint64_t flags,
-								const dnet_time& time_begin, const dnet_time& time_end)
+		uint32_t type, uint64_t flags, const dnet_time& time_begin, const dnet_time& time_end)
```
shaitan: `start_iterator` shouldn't be used with new flags and type, so it is better to check it here and either fail it or unset new flags.

library/net.c
bioothod: Nice catch

library/dnet.c
bioothod: Fixed

example/eblob_backend.c
bioothod: This should actually be `will happen`. I've fixed spelling issues you found

library/dnet.c
bioothod: That's the point, it counts how many bytes are in the wire to any destination group.
There is no need to control source - it can generate as much data as it wants, but it is destination point, that can choke. So there is kind of destination watermark.

library/dnet.c
bioothod: I've copied trace_bit too, but other flags like nolock or direct will break the things.

library/dnet.c
bioothod: I've reworked this chunk a bit to eliminate excessive conversions, it is not needed

library/dnet.c
bioothod: With `DNET_IFLAGS_DATA` flags it can

example/eblob_backend.c
bioothod: No, it can't. It behaves quite differently from how iterator runs, it is more like bulk IO request, the only common thing is response structure sent to client.

example/eblob_backend.c
bioothod: I have no special preference on this - by design server has to sent a response for every key, but if it has failed to send this response to client (because of ENOMEM error), it may proceed with other keys (likely these will also fail), or it may return error early. I think it should return error early, I will add comment about this.

library/dnet.c
bioothod: cmd can not be NULL, check above is being copied from callback to callback. Command structure is either handcrafted in stack when there is error on the client side, or it is stored in transaction, when it is a reply.

library/dnet.c
bioothod: We can't do anything with it - there is no way client can differentiate this particular ENOMEM from other errors with the same error code. I've added response state setting for this error, but more important remove error is not handled here.

It is better to use `dnet_remove_local()` here, it returns error code, but there is no backend control structure in case of server-send command. It is available for iterator request though.

bindings/cpp/session.cpp
bioothod: Actually every flag is allowed here. Flags like range/ts-range will be checked in `dnet_iterator_start()`, data and no-meta flags are also allowed

bindings/cpp/session.cpp
bioothod: It will return an error if there is server-send iterator flag and there are no groups.
New flags are safe for old iterators.

We could clear those flags, but what's the purpose? Blind clearing is wrong - if client has an error and sets up server-send iterator via this call, clearing flags will result in starting usual data iterator, client will think that his data has been copied. I would rather return error for this case. But using other flags is safe.

include/elliptics/packet.h
```diff
@@ -340,30 +359,67 @@ struct dnet_backend_control
 struct dnet_id_container
 {
 	int backends_count;
-	struct dnet_backend_ids backends[0];
 } __attribute__ ((packed));
 
-static inline int dnet_validate_id_container(struct dnet_id_container *ids, size_t size, struct dnet_backend_ids **backends)
+static inline void dnet_convert_id_container(struct dnet_id_container *cnt)
```
shaitan: Why is it a part of API? If it should be a part, why is it changed?

include/elliptics/session.hpp
```diff
@@ -317,7 +317,7 @@ class session
 		 * in this session so they won't collide with same key in any
 		 * other namespace.
 		 */
-		void set_namespace(const std::string &ns);
+		void			set_namespace(const std::string &ns);
```
shaitan: You replaced space by tab, but few lines below you replaced tab by space. Please, use one way for both.

library/dnet.c
```diff
@@ -393,6 +404,384 @@ static int dnet_iterator_callback_send(void *priv, void *data, uint64_t dsize)
...
+					r->data = r->header + sizeof(struct dnet_cmd);
+					r->dsize = sizeof(struct dnet_io_attr);
+					r->st = dnet_state_get(st);
+
+					lc = r->header;
+					dnet_setup_id(&lc->id, send->cmd.id.group_id, cmd->id.id);
+					lc->cmd = DNET_CMD_DEL;
+					lc->backend_id = -1;
+					lc->trace_id = cmd->trace_id;
+					lc->flags = DNET_FLAGS_NOLOCK;
```
shaitan: Remove command should has trace bit flag if it was set in original command (server send or copy iterator).

library/dnet.c
```diff
@@ -393,6 +404,384 @@ static int dnet_iterator_callback_send(void *priv, void *data, uint64_t dsize)
...
+static int dnet_iterator_callback_server_send(void *priv, void *data, uint64_t dsize, int fd, uint64_t data_offset)
+{
+	struct dnet_server_send_ctl *send = priv;
+	struct dnet_net_state *st = send->state;
+	struct dnet_iterator_response *re = data;
+	int err;
+
+	dnet_convert_iterator_response(re);
+
+	err = dnet_server_send_write(send, re, dsize, fd, data_offset);
```
shaitan: This will send write command with unverified data, because iterator doesn't verify checksum.

shaitan: Also this will corrupt keys which are in uncommitted state between 'prepare' and 'commit'. Tests should include cases with uncommitted keys.

library/dnet.c
```diff
@@ -474,13 +866,15 @@ static int dnet_iterator_callback_common(void *priv, struct dnet_raw_id *key, ui
 	response->flags = flags;
 	dnet_convert_iterator_response(response);
 
-	/* Data */
-	err = dnet_read_ll(fd, (char *)position, dsize, data_offset);
-	if (err)
-		goto err_out_exit;
+	if (dsize) {
```
shaitan: It is useless check because `dnet_read_ll` has the same.

library/dnet.c
shaitan: It looks like useless statistics, how do you suppose to use it?

example/eblob_backend.c
shaitan: Why doesn't it use own responses that will not mislead users?

library/dnet.c
shaitan: If `cmd` can not be `NULL` previous check is useless and should be removed.

bindings/cpp/session.cpp
shaitan: `DNET_IFLAGS_NO_META` for this iterator will lead to copying / moving data without meta - it may be kinda strange behaviour, but it is ok and should be mentioned in docs.

shaitan:
This PR extends C++ API and doesn't update Python API, so new features will not be available for python tools. So I suppose there will be another PR with Python API and python tests updates.

include/elliptics/packet.h
bioothod: I reworked id container validation to fix heap overflow and le/be conversion bugs.
It is not used by clients directly, but it is part of structures which client sends to match elliptics protocol, so it lives in public headers.

include/elliptics/session.hpp
bioothod: I made it consistent with the other surrounding methods, you would notice that if opened the whole session class.
Its first methods which do not accept many arguments are all tab-separated, second half of methods which accept larger number of arguments uses only one space.

library/dnet.c
bioothod: Ok. NULL check above has also been removed

library/dnet.c
bioothod: The same logic applies to any other iterator type - they neither check checksums nor deal with uncommitted keys.
I'm not going to change it in this pull request. I would recommend performing checksum in common iterator callback

include/elliptics/session.hpp
shaitan: Let's make all consistent and will use one space for all methods and tabs only for members?

library/dnet.c
bioothod: Relying on knowing internal structure of the read call is not clear. When there is explicit size check reader knows everything is safe.

library/dnet.c
bioothod: This is a flow control, which does not allow overflowing of the destination node, which will be busy writing data to the low-level storage.

example/eblob_backend.c
bioothod: I'm the user and it doesn't mislead me. It is not an iterator, it behaves in a quite different way, but underlying idea is similar.
Iterator runs over key range and performs actions on every key found, this API provides list of keys to run over and perform that same set of actions. Thus reply is the same. The only difference is iterator ID since there is no entity to stop when DNET_CMD_SEND is used - all keys have been already transferred to multiple server nodes and server's IO threads started to read/send the data.

library/dnet.c
shaitan: Common iterator doesn't check checksum and doesn't deal with uncommitted keys, because it was designed for recovery which deals with uncommitted keys and reads data with checksum verification. Performing checksum verification in all iterator callback will decrease speed of iteration. Iterator could verify checksum only if it is run with `DNET_IFLAGS_DATA` and/or for new iterator type.

bindings/cpp/session.cpp
bioothod: It is exactly the same logic as with all other iterator types - when no-meta flag is set no metadata is transferred either to client or to remote node. Consider server-send iterator as a special type of usual data iterator where 'client' is some another server node.

library/dnet.c
shaitan: Relying on knowing internal structure of the read call allows to escape this check in all places where it is used.

library/dnet.c
bioothod: Verifying checksum might not that good idea actually - there might be the case that server runs with no-checksum flag, or it will run with this flag, but iterator has failed to copy data or to send it to client for recovery. This ends up with some keys have not been recovered although they can be read after no-checksum flag has been turned on.
Uncommitted keys raise different issue - send iterator writes them as completed writes, this is probably not what is expected. Data iterator for instance returns it too, but recovery will skip it, since it will read the key and it will fail.
I will add a check which skips uncommitted data and add this into documentation.

This is also a problem with current code - reading data will fail, recovery will skip the key, it will not be copied, turning off checksum verification *returns* key back.

I'm not sure whether this is a real problem or not, I would rather forbid writes without checksums.

bioothod:
I can add this API call, but I'm not sure what tests you want? It is the same C++ `session::start_copy_iterator()` call wrapped into python's `create_result()`.

I would also recommend using it in recovery, but I do not have strong opinion on that and will not change recovery code.

library/dnet.c
```diff
@@ -393,6 +404,384 @@ static int dnet_iterator_callback_send(void *priv, void *data, uint64_t dsize)
...
+
+	dnet_convert_iterator_response(re);
+
+	err = dnet_server_send_write(send, re, dsize, fd, data_offset);
+
+	if (atomic_read(&send->bytes_pending) > DNET_SERVER_SEND_WATERMARK_HIGH) {
+		pthread_mutex_lock(&send->write_lock);
+		while ((atomic_read(&send->bytes_pending) > DNET_SERVER_SEND_WATERMARK_LOW) &&
+				!st->__need_exit && !send->write_error) {
+			if (!st->__need_exit)
```
abudnik: I think this condition is unnecessary.

bioothod: Makes sense

library/dnet.c
bioothod: This is the only place where it is used

shaitan: So this is the only place where this useless check can be escaped.

include/elliptics/packet.h
shaitan: It is dark part of protocol which isn't available for client use, so living it in public headers is useless.

example/eblob_backend.c
shaitan: You are not representative user. This will confuse the most of users who didn't examine Elliptics source code.

shaitan:
Recovery will use these features but with restriction that it doesn't copies/moves big records and doesn't deal with records' timestamp.

bioothod:
Copying big files with this approach is much safer and faster.
I introduced prepare/commit commands to handle completely, absolutely different problem.

In the server life there are moments when 100 people slowly upload 1Gb files. This may eat 100Gb of RAM if handled naively. Thus we have prepare/commit and chunked writes.

Server-send operation is very different by its nature. There are no 100 clients who might copy 1Gb files, instead there is a one client who copies files one after another with limited number of writes in flight, currently it is limited to 30Gb. The only possible problem is the fact that there are files which are larger than RAM. We do not have such files. You do not have such files (I suspect you do not have files more than a couple of gigs). There is no problem with copying large files with this approach.

And instead of making recovery 2 times faster for free, instead of removing need to run recovery script on the server, which in turn requires ssh or running special agents - all this complex things - you make what? Arguing that there is zero probability of what?

This new code has to be used in recovery. It was designed exactly for that purpose.
Even more, it is already used in alternative implementation, and it proved it works great.

example/eblob_backend.c
bioothod: I'm the only one user of this feature so far.
And this is how it is designed. Because it is very convenient.

It. Is. Very. Convenient.

You may think whatever you like, but you didn't even try to use it or to think why is it so.
You would read test code to check how simple became handling of the server-side copy process compared to iterator code. You haven't.

This whole "review" process becomes useless. Instead of searching for bugs or trying to understand what it is and how it should be used, we argue about what? That is has to be done in a different way because of why?
Because it will mislead users? This is not an argument at all. This is not a review. Kirill, please reconsider what are you doing here now.

library/dnet.c
bioothod: The real data corruption bug was introduced because this check was removed.
Originally code was safe and then *optimized* to hide this check, this optimization removed correct offset calculation, which ended up with overwriting iterator response with data. Apparently you do not use this code at all.

I fixed this bug and returned back explicit check.

include/elliptics/packet.h
bioothod: I put every command and data structure which is part of the protocol into public headers because one might want to implement Elliptics client in a different language. No one tried so far, but it does not matter.

This pull request does not and will not change how our commands and structures are stored in the source code.
If you think that half of them has to go into private headers, we can discuss it, most likely I will not accept it because of the reason above. And this whole pull request is not about it. Even this chunk is not about it, it fixes 2 bugs - le/be conversion and possible memory corruption. This could not be done without changing container parser function, and this bug was only visible in server-send test, since it starts servers in parallel and immediately starts working with several groups.

include/elliptics/session.hpp
bioothod: Ok, I've updated session class

shaitan:
I wasn't arguing with you at all, I just expressed my point of view and I tired of your aggression. This PR showed that we have to change the way how we cook, describe and review pull requests.

If you have nothing to change within this PR, it can be merged.

bioothod:
Yes, this discussion went on higher tones. I have to apologize and I really do if I somehow insulted you.

But discussion is so much far away from the original purpose, so much into nitpicking of the stuff that is not meaningful at all instead of concentrating of the core, that things get exploded. Expressions like 'I tired' do not help either, instead we would try to come up with the purpose of the review, to remember why is it needed at all. Again, if I insulted you, I apologize.

In my opinion review has to deal with bugs and with attempts to understand how things work. We discuss in advance design of the system and rough implementation, it is very wrong to show the code before showing the idea. This is not the case with this request. It is also not the place to try to show the power or *portering* like 'he fucks every other pull request, let's tear his one apart'. We have to be fair and clear equally to ourselfs and peers.

When things get so much far away from the *discussion* of the idea, from the talk, dialogue into definitive terms like 'this is wrong' without even trying to understand why is it so, I do not see any other way like replying with more expression. I'm not sure this is the always correct way to handle this situation though.

I do hope this will not break the relationships we have, but will help focusing on the right things.
Please be more detailed with your concerns, let's try not to use definitive terms in the questionable contexts and respect what others do.

PR #652, opened 2015-09-18
config: read_only option for backend

abudnik:
read_only option for backend makes backend read only at startup.

### DIFF
--- a/example/config.cpp
+++ b/example/config.cpp
@@ -332,6 +332,7 @@ void parse_backends(config_data *data, const config &backends)
 
 		dnet_backend_info &info = backends_info[backend_id];
 		info.enable_at_start = backend.at<bool>("enable", true);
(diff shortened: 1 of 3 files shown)


bioothod:
All checks have failed, should it be merged?

abudnik:
All checks failed, because elliptics must depend on newer eblob version (compiler couldn't find chunks_dir variable in eblob_config struct).
PR could be merged.

example/config.cpp
```diff
@@ -332,6 +332,7 @@ void parse_backends(config_data *data, const config &backends)
 
 		dnet_backend_info &info = backends_info[backend_id];
 		info.enable_at_start = backend.at<bool>("enable", true);
+		info.read_only_at_start = backend.at<bool>("read_only", false);
```
shaitan: This should be added to `example/ioserv.json` and described in docs.

PR #653, opened 2015-09-23
cleanup: stop all node thread pools, then destroy monitor, then clean…

abudnik:
…up node thread pools

If pool threads are not stopped before cleaning monitor object, then some requests may use already destroyed monitor.

### DIFF
--- a/library/server.c
+++ b/library/server.c
@@ -220,17 +220,17 @@ void dnet_server_node_destroy(struct dnet_node *n)
 	dnet_log(n, DNET_LOG_DEBUG, "Destroying server node.");
 
 	/*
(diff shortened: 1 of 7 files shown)


library/server.c
```diff
@@ -219,6 +219,8 @@ void dnet_server_node_destroy(struct dnet_node *n)
 {
 	dnet_log(n, DNET_LOG_DEBUG, "Destroying server node.");
 
+	dnet_node_stop_common_resources(n);
+
```
bioothod: This is a contradictory to what is written in comments about monitor below.
Please update both monitor comments and add description of the new cleanup logic.

abudnik: fixed

PR #657, opened 2015-10-15
Get rid of unused backends

bioothod:
We do not extend and add features into file and module backend.
Also there are no known users of those backends.


shaitan:
Please, remove `usr/lib/libelliptics_module_backend_cpp.so.*` from [debian/elliptics.install](https://github.com/reverbrain/elliptics/blob/master/debian/elliptics.install#L3)

bioothod:
Fixed, thank you.

shaitan:
Looks good, it can be merged. Thank you!

PR #658, opened 2015-10-15
srw: added ability to reply via response stream

toshic:


srw/srw.cpp
```diff
@@ -74,17 +89,60 @@ class dnet_upstream_t: public cocaine::api::stream_t
 		}
 
 		virtual void write(const char *chunk, size_t size) {
-			reply(false, chunk, size);
```
bioothod: Please add extensive comments describing the new logic

srw/srw.cpp
```diff
@@ -74,17 +89,60 @@ class dnet_upstream_t: public cocaine::api::stream_t
...
+			if (rv != msgpack::UNPACK_SUCCESS && rv != msgpack::UNPACK_EXTRA_BYTES) {
+				err = EINVAL;
+				SRW_LOG(*m_node->log, DNET_LOG_ERROR, "app/" + m_sph.event, "unable to unpack response");
+			} else if (obj.type != msgpack::type::RAW) {
+				err = EINVAL;
+				SRW_LOG(*m_node->log, DNET_LOG_ERROR, "app/" + m_sph.event, "insorrect response type, should be string, got: %d", obj.type);
+			}
+
+			if (err) {
+				error(err, "protocol error");
```
bioothod: Should it be more descriptive message?

Also, why does it use SRW_LOG directly and error() wrapper simultaneously?

srw/srw.cpp
```diff
@@ -74,17 +89,60 @@ class dnet_upstream_t: public cocaine::api::stream_t
...
+				SRW_LOG(*m_node->log, DNET_LOG_ERROR, "app/" + m_sph.event, "unable to allocate memory: %d", errno);
+				return;
+			}
+
+			/* Make SPH reply for client */
+			char *p = data;
+			sph * sph_p = (sph*)data;
+
+			memcpy(data, &m_sph.sph, sizeof(sph));
+			p += sizeof(sph);
```
bioothod: Should it check all data sizes?

srw/srw.cpp
```diff
@@ -74,17 +89,60 @@ class dnet_upstream_t: public cocaine::api::stream_t
...
+
+			sph_p->event_size = m_sph.event.size();
+			memcpy(p, m_sph.event.data(), m_sph.event.size());
+			p += m_sph.event.size();
+
+			sph_p->data_size = obj.via.raw.size;
+			memcpy(p, obj.via.raw.ptr, obj.via.raw.size);
+
+			reply(false, data, data_size);
+			free(data);
```
bioothod: Is it possible to catch some exception in the middle of the function?

srw/srw.cpp
```diff
@@ -479,6 +541,7 @@ class srw {
 					}
 
 					stream->write((const char *)sph, total_size(sph) + sizeof(struct sph));
+					stream->close();
```
bioothod: Why is it needed? Looks like it is changing the behaviour

srw/srw.cpp
```diff
@@ -506,6 +569,24 @@ class srw {
 			return err;
 		}
 
+		void complete_job(std::string id, sph_wrapper sph)
+		{
+			char sph_str[DNET_DUMP_NUM * 2 + 1];
+			dnet_dump_id_len_raw(sph.sph.src.id, DNET_DUMP_NUM, sph_str);
+			sph_str[2 * DNET_DUMP_NUM] = '\0';
```
bioothod: Please move it into error path.
Also, you can use `dnet_dump_id_str()`, it is thread-safe

srw/srw.cpp
bioothod: It also looks like need new tests

srw/srw.cpp
toshic: Didn't get what do you want to say. event_size and data_size will be overwritten a bit later, during copy.

srw/srw.cpp
toshic: All functins inside this block don't throw exceptions.

srw/srw.cpp
toshic: It's request stream. Here we indicate that there will be no more data sent to worker.

srw/srw.cpp
bioothod: It looks like data_size calculation already includes all structures it is about to host. My question was about the case whether or not it is enough space to host them all.

srw/srw.cpp
bioothod: `reply()` can raise an exception

srw/srw.cpp
```diff
@@ -458,7 +517,10 @@ class srw {
 
 				it->second->update(event, sph);
 
-				dnet_shared_upstream_t upstream(std::make_shared<dnet_upstream_t>(m_node, st, cmd, event, (uint64_t)sph->flags));
+				dnet_shared_upstream_t upstream(std::make_shared<dnet_upstream_t>(
+                                                    m_node, st, cmd, sph,
+                                                    std::bind(&srw::complete_job, this, id_str, sph_wrapper(sph))
+												));
```
bioothod: Indent is heavily broken here

srw/srw.cpp
bioothod: Why we didn't it before? What has changed from worker's and client perspective with this change?

srw/srw.cpp
```diff
@@ -458,7 +521,8 @@ class srw {
 
 				it->second->update(event, sph);
 
-				dnet_shared_upstream_t upstream(std::make_shared<dnet_upstream_t>(m_node, st, cmd, event, (uint64_t)sph->flags));
+				dnet_shared_upstream_t upstream(std::make_shared<dnet_upstream_t>( m_node, st, cmd, sph,
```
bioothod: Indent is broken at `std::make_shared<dnet_upstream_t>( m_node` - a space after `( `

PR #659, opened 2015-10-16
Timestamp cas

bioothod:
New IO flag which forbids overwriting data if its on-disk (or cached) timestamp is newer than that of data about to be written.
It is tested in server-send suite, by default server-send uses this flag and thus test where we write data with old timestamp and then try to move it to different groups where the same keys exist with newer timestamp. In this operation move should fail, i.e. all keys should not be moved, which implies keys will not be locally removed, and that's exactly what is tested.

Clearing timestamp cas bit (by setting iterator's overwrite flag) forces test to overwrite data, checking new content should succeed.

### DIFF
--- a/example/eblob_backend.c
+++ b/example/eblob_backend.c
@@ -180,70 +180,6 @@ static int blob_lookup(struct eblob_backend *b, struct eblob_key *key, struct eb
 	return err;
 }
 
-static int blob_cas_timestamp(struct eblob_backend_config *c, struct eblob_key *key, struct dnet_io_attr *io)
-{
-	struct dnet_ext_list_hdr ehdr;
-	struct dnet_ext_list elist;
-	struct eblob_backend *b = c->eblob;
-	struct eblob_write_control wc = { .data_fd = -1 };
-	int err;
-
-	dnet_ext_list_init(&elist);
-
-	err = blob_lookup(b, key, &wc);
-	if (err < 0) {
-		if (err != -ENOENT) {
-			dnet_backend_log(c->blog, DNET_LOG_ERROR, "%s: EBLOB: blob-cas-timestamp: LOOKUP: %d: %s",
-				 dnet_dump_id_str(io->id), err, strerror(-err));
-			goto err_out_exit;
-		}
-
-		/* There is no record, write is safe. */
-		err = 0;
-		goto err_out_exit;
-	}
-
-	/* If it is data with extended header, we must check timestamp.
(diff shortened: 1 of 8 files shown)


example/eblob_backend.c
```diff
@@ -214,6 +278,16 @@ static int blob_write(struct eblob_backend_config *c, void *state,
 
 	memcpy(key.id, io->id, EBLOB_ID_SIZE);
 
+	/*
+	 * Check timestamp of the existing record if compare-and-swap timestamp flag is set.
+	 * If on-disk record's timestamp is higher than that in IO structure, do not allow this write.
+	 */
+	if (io->flags & DNET_IO_FLAGS_CAS_TIMESTAMP) {
```
shaitan: Why this check is made here, not in `dnet_cas_local()`?

tests/server_send.cpp
```diff
@@ -77,6 +77,28 @@ static void ssend_configure(const std::string &path)
 	ssend_servers = tests::start_nodes(cfg);
 }
 
+static void ssend_test_insert_many_keys_old_ts(session &s, int num, const std::string &id_prefix, const std::string &data_prefix)
+{
+	s.set_trace_id(rand());
+	for (int i = 0; i < num; ++i) {
+		std::string id = id_prefix + lexical_cast(i);
```
abudnik: lexical_cast(i) -> std::to_string(static_cast<long long int>(i))

bioothod: Is it available in all supported compilers? Also, why is `static_cast()` needed?
`boost::lexical_cast()` (which is much slower than our version) is heavily used in tests, it would be great to use one common way for it

example/eblob_backend.c
bioothod: We do not have API to get neither file info nor `dnet_io_attr` from the backend.
To optimize things, backend sends requested data to the client directly, higher layer like `dnet_process_cmd_raw()` and IO threads can not get this info.

tests/server_send.cpp
abudnik: the only available specialization of to_string is std::to_string<long long int> on gcc 4.4 (lucid)

bioothod: Well, makes sense to replace all `lexical_cast()` calls (boost and ours) with `std::to_string()` as a separate refactoring pull request

shaitan:
:+1: Looks good, thank you!

PR #661, opened 2015-11-05
Returned ability to enable uninitialized backend if its config is presented

shaitan:


library/backend.cpp
```diff
@@ -645,9 +646,21 @@ static int dnet_cmd_backend_control_dangerous(struct dnet_net_state *st, struct
 
 	const dnet_backend_info &backend = backends[control->backend_id];
 	if (backend.state == DNET_BACKEND_UNITIALIZED) {
-		dnet_log(node, DNET_LOG_ERROR, "backend_control: there is no such backend: %u, state: %s",
-				control->backend_id, dnet_state_dump_addr(st));
-		return -EINVAL;
+		switch(dnet_backend_command(control->command)) {
+			/* uninitialized backend can be enabled */
+			case DNET_BACKEND_ENABLE:
```
bioothod: What is that?

This is quite ambiguous and misleading to enable uninitialized backend.
It is like writing into uninitialized array, which leads to bugs.

shaitan: On startup, Elliptics marks all backends without config as uninitialized. These changes allow to enable backends which config were added.

bioothod: What if nothing was added into config?
How can one distinguish uninitialized backend from disabled one?

shaitan: >What if nothing was added into config?

If nothing was added it will fail enabling.

>How can one distinguish uninitialized backend from disabled one?

By state, uninitialized backend has state `-1` (`DNET_BACKEND_UNITIALIZED`) and disabled - `0` (`DNET_BACKEND_DISABLED`).

bioothod: What's wrong with `"enable": false` option in config?
Will be uninitialized or disabled? I suppose disabled, since it can be started later.

So what's the purpose of adding holes in backend enumeration and then add hacks which look like `write into uninitialized array`?

shaitan:
We decided to change it - users should have ability to add backends without using hidden feature of uninitialized backends.

PR #662, opened 2015-11-07
Ioflags mix states

bioothod:
New IO flag which forces client to mix states according to their weights.
Sometimes it is desired to only mix states for some requests and use strict order for others, this new IO flags allows just that.
Only created for weighted mix and not randomization, since the latter can be achieved in client.

### DIFF
--- a/bindings/python/elliptics_python.cpp
+++ b/bindings/python/elliptics_python.cpp
@@ -59,6 +59,7 @@ enum elliptics_cflags {
 
 enum elliptics_ioflags {
 	ioflags_default			= 0,
+	ioflags_mix_states		= DNET_IO_FLAGS_MIX_STATES,
 	ioflags_append			= DNET_IO_FLAGS_APPEND,
 	ioflags_prepare			= DNET_IO_FLAGS_PREPARE,
 	ioflags_commit			= DNET_IO_FLAGS_COMMIT,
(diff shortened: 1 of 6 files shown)


shaitan:
Since `DNET_IO_FLAGS_MIX_STATES` and `DNET_IO_FLAGS_SKIP_SENDING` are the same, will server node send anything if client send a command with `DNET_IO_FLAGS_MIX_STATES`?

This should be covered by tests, please add them.

bindings/python/elliptics_python.cpp
```diff
@@ -421,6 +422,7 @@ BOOST_PYTHON_MODULE(core)
 		"cache_remove_from_disk\n    is set and object is being removed from cache,\n"
 		                        "    then remove object from disk too")
 		.value("default", ioflags_default)
+		.value("mix_states", ioflags_mix_states)
```
shaitan: Since you introduce new flag - please, add description for it.

bioothod:
Actually, I've found that skip-sending bit can not be shared with any IO command, since eblob backend reuses IO attribute structure and thus mix-states bit becomes skip-sending, which prevents read reply from being sent to client.

I've found that by running a new crafted test for that bit, thanks for catching this.

Since skip-sending bit can not be reused, I've allocated new bit for this flag.
Python docs and test have been updated, I've also added `cas_timestamp` flag and description.

shaitan:
Thank you!

PR #663, opened 2015-11-07
iterator: automatically set key range flag when ranges are present

bioothod:
What about automatically setting iterator flags in cpp binding if there are ranges?


shaitan:
Yes, it can be done in cpp binding too.

bioothod:
I've update `iterator()` helper, it sets key-range flag automatically now if there is at least one range in the request

shaitan:
Should it be covered by test that will guarantee that it won't be broken?

bioothod:
I'm not sure whether this should be tested at all, after all, nothing explodes if we continue using old behavior with explicit flags setting.
I've slightly modified single-range-iterator test though, where dropped flags settings, this test should not fail. I also added comments to reflect that flags are set explicitly.

shaitan:
Thank you

PR #667, opened 2015-11-18
Fixed server_send() aggregation error

bioothod:
After request has been completed, final ack will be processed both by common callback and iterator callback checkers. The latter will fail, since final ack doesn't bring any data.
This is generally worked around by using clean session clones (without checkers) and forcing aggregator to only work with clean session, i.e. apply all checkers on final aggregated results, not at intermediate replies.

### DIFF
--- a/bindings/cpp/callback.cpp
+++ b/bindings/cpp/callback.cpp
@@ -68,6 +68,9 @@ public:
 		return false;
 	}
 
+	// how many independent transactions share this handler plus call below
+	// call below and corresponding +1 is needed, since transactions can be completed
+	// before send_impl() calls this method to setup this 'reference counter'
 	bool set_total(size_t total)
 	{
 		m_handler.set_total(total);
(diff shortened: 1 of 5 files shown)


shaitan:
It should also include updates to the tests. For example:
* `server_send` tests shouldn't set exception policy to `no_exceptions` for all commands except those expected to fail
* it should be tested that `server_send` throws an exception when some or all keys were failed to send due to timeout, missing or corrupted keys
* it should be tested that filters, checkers and exception policies works correctly for `server_send`

bioothod:
You can not enable exceptions since there are tests which deliberately wait for error to check that under some conditions operation fails.

This same issue exists for iterators and likely bulk operations too. It is not really possible to test for this error after it has been fixed. Its like testing for non-existing file when you do know that you've written that file successfully.

The only feasible test I see is to check errors while reading server-send/copy-iterator responses, I added some of these checks into copy iterator test, server-send should have them too.

bioothod:
Actually we already do this implicitly, since we test both server-send and copy iterators by reading every key they are supposed to copy, and if reading fails so does the test.

We can simplify this by checking every returned key status during copy and not after copy has been completed. This might catch the error, when keys have been successfully copied, but still iterator returned one final error, which was the reason for this pull request.

bioothod:
Commit 75289c8 should do exactly that.

PR #668, opened 2015-11-19
dc_recovery: optimization: use server-send for small/medium keys

abudnik:
For detailed information see recovery/recovery.qdoc

This optimization speed up in many times dc recovery process for small/medium keys (< chunk-size == 1Mb by default).

### DIFF
--- a/recovery/elliptics_recovery/dc_recovery.py
+++ b/recovery/elliptics_recovery/dc_recovery.py
@@ -22,6 +22,7 @@ import errno
 
 from elliptics_recovery.utils.misc import elliptics_create_node, RecoverStat, validate_index
 from elliptics_recovery.utils.misc import INDEX_MAGIC_NUMBER_LENGTH, load_key_data, WindowedRecovery
+from elliptics_recovery.dc_server_send import ServerSendRecovery
 
 import elliptics
 from elliptics import Address
@@ -383,7 +384,8 @@ class WindowedDC(WindowedRecovery):
     def __init__(self, ctx, node):
         super(WindowedDC, self).__init__(ctx, ctx.stats['recover'])
         self.node = node
-        self.keys = iterate_key(self.ctx.merged_filename, self.ctx.groups)
+        ctx.rest_file.flush()
+        self.keys = iterate_key(self.ctx.rest_file.name, self.ctx.groups)
 
     def run_one(self):
         try:
@@ -402,6 +404,14 @@ class WindowedDC(WindowedRecovery):
         return False
 
 
+def cleanup(ctx):
+    for f in ctx.bucket_files.itervalues():
+        os.remove(f.name)
+    del ctx.bucket_files
+    os.remove(ctx.rest_file.name)
+    del ctx.rest_file
+
+
 def recover(ctx):
     stats = ctx.stats['recover']
 
@@ -412,7 +422,9 @@ def recover(ctx):
                                  net_thread_num=4,
                                  io_thread_num=24,
                                  remotes=ctx.remotes)
-    result = WindowedDC(ctx, node).run()
+    result = ServerSendRecovery(ctx, node).recover()
+    result &= WindowedDC(ctx, node).run()
+    cleanup(ctx)
     stats.timer('recover', 'finished')
 
     return result
--- a/recovery/elliptics_recovery/dc_server_send.py
+++ b/recovery/elliptics_recovery/dc_server_send.py
@@ -144,12 +144,17 @@ class ServerSendRecovery(object):
         log.info("Server-send bucket: source group: {0}, num keys: {1}".format(group, len(keys)))
         keys_bunch = dict() # remote_groups -> [list of newest keys]
         for key, key_infos in keys:
-            filtered_key_infos = self._get_unprocessed_key_infos(key_infos, group)
-            if not self._can_use_server_send(filtered_key_infos):
+            unprocessed_key_infos = self._get_unprocessed_key_infos(key_infos, group)
+
+            is_first_attempt = len(unprocessed_key_infos) == len(key_infos)
+            if is_first_attempt and self._process_uncommited_keys(key, key_infos):
+                continue
+
+            if not self._can_use_server_send(unprocessed_key_infos):
                 self.buckets.move_to_rest_bucket(key, key_infos)
                 continue
 
-            dest_groups = self._get_dest_groups(filtered_key_infos)
+            dest_groups = self._get_dest_groups(unprocessed_key_infos)
             index = frozenset(dest_groups)
             if index not in keys_bunch:
                 keys_bunch[index] = []
@@ -176,8 +181,8 @@ class ServerSendRecovery(object):
                     iterator = self.session.server_send(newest_keys, 0, remote_groups)
                     timeouted_keys, corrupted_keys = self._check_server_send_results(iterator, key_infos_map, group)
                     newest_keys = timeouted_keys
-                    if corrupted_keys and not self.ctx.safe:
-                        self._remove_corrupted_keys(corrupted_keys, group)
+                    if corrupted_keys:
+                        self._remove_corrupted_keys(corrupted_keys, [group])
 
             if timeouted_keys:
                 self._on_server_send_timeout(timeouted_keys, key_infos_map, group)
@@ -237,15 +242,18 @@ class ServerSendRecovery(object):
             next_group = self._get_next_group(key_infos, group)
             self.buckets.on_server_send_fail(key, key_infos, next_group)
 
-    def _remove_corrupted_keys(self, keys, group):
+    def _remove_corrupted_keys(self, keys, groups):
         '''
         Removes invalid keys with invalid checksum.
         '''
-        self.remove_session.set_groups([group])
+        if self.ctx.safe:
+            return
 
         for attempt in range(self.ctx.attempts):
             if not keys:
                 break
+
+            self.remove_session.set_groups(groups)
 
             results = []
             for k in keys:
@@ -253,13 +261,54 @@ class ServerSendRecovery(object):
                 results.append(result)
 
             timeouted_keys = []
+            timeouted_groups = set()
             is_last_attempt = (attempt == self.ctx.attempts - 1)
             for i, r in enumerate(results):
                 status = r.get()[0].status
                 log.info("Removing corrupted key: {0}, status: {1}, last attempt: {2}".format(keys[i], status, is_last_attempt))
                 if status == -errno.ETIMEDOUT:
                     timeouted_keys.append(keys[i])
-            keys = timeouted_keys
+                    timeouted_groups.add(keys[i].group_id)
+            keys, groups = timeouted_keys, list(timeouted_groups)
+
+    def _process_uncommited_keys(self, key, key_infos):
+        same_ts = lambda lhs, rhs: lhs.timestamp == rhs.timestamp
+        same_infos = [info for info in key_infos if same_ts(info, key_infos[0])]
+
+        same_uncommitted = [info for info in same_infos if info.flags & elliptics.record_flags.uncommitted]
+        has_uncommitted = len(same_uncommitted) > 0
+        if same_uncommitted == same_infos:
+            # if all such keys have exceeded prepare timeout - remove all replicas
+            # else skip recovering because the key is under writing and can be committed in nearest future.
+            same_groups = [info.group_id for info in same_infos]
+            if same_infos[0].timestamp < self.ctx.prepare_timeout:
+                groups = [info.group_id for info in key_infos]
+                log.info('Key: {0} replicas with newest timestamp: {1} from groups: {2} are uncommitted. '
+                         'Prepare timeout: {3} was exceeded. Remove all replicas from groups: {4}'
+                         .format(key, same_infos[0].timestamp, same_groups, self.ctx.prepare_timeout, groups))
+                self._remove_corrupted_keys([key], groups)
+            else:
+                log.info('Key: {0} replicas with newest timestamp: {1} from groups: {2} are uncommitted. '
+                         'Prepare timeout: {3} was not exceeded. The key is written now. Skip it.'
+                         .format(key, same_infos[0].timestamp, same_groups, self.ctx.prepare_timeout))
+            return True
+        elif has_uncommitted:
+            # removed incomplete replicas meta from same_infos
+            # they will be corresponding as different and will be overwritten
+            same_infos = [info for info in same_infos if info not in same_uncommitted]
+            incomplete_groups = [info.group_id for info in same_uncommitted]
+            same_groups = [info.group_id for info in same_infos]
+            log.info('Key: {0} has uncommitted replicas in groups: {1} and completed replicas in groups: {2}.'
+                     'The key will be recovered at groups with uncommitted replicas too.'
+                     .format(key, incomplete_groups, same_groups))
+
+            committed_infos = [info for info in key_infos if info not in same_uncommitted]
+            if committed_infos:
+                key_infos = same_uncommitted + committed_infos
+                next_group = committed_infos[0].group_id
+                self.buckets.on_server_send_fail(key, key_infos, next_group)
+
+        return has_uncommitted
 
     def _get_dest_groups(self, key_infos):
         '''
@@ -288,13 +337,7 @@ class ServerSendRecovery(object):
         return []
 
     def _can_use_server_send(self, key_infos):
-        return not self._check_key_chunked(key_infos) and not self._check_key_uncomitted(key_infos)
-
-    def _check_key_uncomitted(self, key_infos):
-        return key_infos[0].flags & elliptics.record_flags.uncommitted
-
-    def _check_key_chunked(self, key_infos):
-        return key_infos[0].size > self.ctx.chunk_size
+        return key_infos[0].size < self.ctx.chunk_size
 
     def _get_next_group(self, key_infos, group):
         '''
--- a/recovery/elliptics_recovery/dc_server_send.py
+++ b/recovery/elliptics_recovery/dc_server_send.py
@@ -1,0 +1,274 @@
+import logging
+import os
+import errno
+import time
+from itertools import groupby
+
+from elliptics_recovery.utils.misc import dump_key_data, load_key_data_from_file
+
+import elliptics
+
+log = logging.getLogger()
+
+
+class BucketsManager(object):
+    '''
+    BucketsManager is a single repository of bucket files. It is used by ServerSendRecovery.
+    '''
+    def __init__(self, ctx):
+        self.ctx = ctx
+        self.bucket_index = -1
+        self.buckets = dict()
+
+    def get_next_bucket(self):
+        '''
+        Every call returns BucketKeys object in a round-robin manner.
+        '''
+        if not self.ctx.bucket_order:
+            return None
+
+        self.bucket_index = (self.bucket_index + 1) % len(self.ctx.bucket_order)
+        group = self.ctx.bucket_order[self.bucket_index]
+        log.debug("Get next bucket: index: {0}, group: {1}, bucket_order: {2}".format(self.bucket_index, group, self.ctx.bucket_order))
+        return self._get_bucket(group)
+
+    def on_server_send_fail(self, key, key_infos, next_group):
+        if next_group >= 0:
+            b = self._get_bucket(next_group)
+            b.add_key(key, key_infos)
+        else:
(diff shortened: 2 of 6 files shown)


agend:
Is it save to  use? We actually need to add several nodes to our cluster and we can use it and test how it will go. Make sense?

recovery/elliptics_recovery/dc_server_send.py
```diff
@@ -0,0 +1,206 @@
...
+import os
+import errno
+import time
+
+from elliptics_recovery.utils.misc import dump_key_data, load_key_data_from_file
+
+import elliptics
+
+log = logging.getLogger()
+
```
bioothod: Please add documentation for these new classes and methods, it is really hard to follow this new recovery mechanism without understanding what new classes are needed for

recovery/elliptics_recovery/dc_server_send.py
```diff
@@ -0,0 +1,206 @@
...
+    def __init__(self, ctx, node):
+        self.ctx = ctx
+        self.stats = ctx.stats['recover']
+        self.node = node
+        self.buckets = BucketsManager(ctx)
+
+        self.session = elliptics.Session(node)
+        self.session.trace_id = ctx.trace_id
+        self.session.exceptions_policy = elliptics.exceptions_policy.no_exceptions
+        self.session.timeout = 120 # magic ?
```
bioothod: Why is it magic?

120 seconds should probably be large enough to wait for the first iterator result since it may skip many keys, but in server-send case results should be received very quickly after beginning of the call

abudnik: I've just copied this value from tests/server_send.cpp.
So, is it necessary to set this timeout to 120?

bioothod: I do not remember default timeout, is it 30 seconds?
This is not needed for `server_send()` method, but may be needed for iterator especially if it requires metadata

abudnik:
```
$> dnet_recovery --help
...
  -w WAIT_TIMEOUT, --wait-timeout=WAIT_TIMEOUT
                        [Wait timeout for elliptics operations default: 3600]
```

I will use default and customizable wait-timeout value.

bioothod: This was a timeout for the whole operation, such huge values are not required anymore.
Basically, this timeout should reflect time for iterator to send next item, it may take a while if we have ranges - iterator will skip keys and client may reset connection waiting for the next one; and also this timeout should reflect potentially random reads and writes.

shaitan:
@agend This PR includes changes only in dc recovery (replicas synchronization), but you need merge recovery which will redistribute data in the cluster after adding new nodes. Changes in merge recovery is on the way and will be submitted in a separate PR.

agend:
@shaitan Cool! Do you have ETA for it? We can really test it)

shaitan:
@agend I think PR with merge recovery will be submitted next week, so you will can taste it and we would appreciate if you share your experience of using it.

recovery/elliptics_recovery/dc_server_send.py
shaitan: Default `wait_timeout` is 5 seconds, 30 seconds - is the default `check_timeout`. `server_send` can use timeout specified by `-w WAIT_TIMEOUT`, but its default should be decreased to some empirical value (`3600` was a upper bound which should never be achieved).

recovery/elliptics_recovery/dc_server_send.py
abudnik: https://github.com/abudnik/elliptics/commit/78a769277baebf59e33bc1bf0f2a3bdaae871a80

bioothod:
Does it need new tests or the old ones are enough?

recovery/elliptics_recovery/utils/misc.py
```diff
@@ -327,6 +327,13 @@ def load_key_data(filepath):
             yield (elliptics.Id(data[0], 0), tuple(KeyInfo.load(d) for d in data[1]))
 
 
+def load_key_data_from_file(keys_file):
```
shaitan: It repeats `load_key_data`. Why do you need both of them?

recovery/recovery.qdoc
```diff
@@ -203,7 +207,19 @@
...
-  Built-in implementation of custom recovery for each key from merged results:\n
+  Built-in implementation of custom recovery for each key from merged results consists of two steps: server-send recovery with following rest-keys recovery.\n
+  If key could not be recovered via server-send, then it will be processed during rest-keys recovery.\n
+
+  Server-send recovery:
+  -# while there are non-empty buckets:
+    -# get next bucket - file that contains newest keys from a specific group
+    -# iterate over bucket keys:
+      -# check that key could be recovered via server-send, if so:
+        -# get destination groups to which key must be recovered and call server-send for it
```
shaitan: According to this, recovery calls server-send for each keys. Doesn't it combine keys into batches and call server-send for each batches, does it?

recovery/recovery.qdoc
```diff
@@ -203,7 +207,19 @@
...
 
-  Built-in implementation of custom recovery for each key from merged results:\n
+  Built-in implementation of custom recovery for each key from merged results consists of two steps: server-send recovery with following rest-keys recovery.\n
+  If key could not be recovered via server-send, then it will be processed during rest-keys recovery.\n
+
+  Server-send recovery:
+  -# while there are non-empty buckets:
+    -# get next bucket - file that contains newest keys from a specific group
+    -# iterate over bucket keys:
+      -# check that key could be recovered via server-send, if so:
```
shaitan: It should be described how recovery checks this.

recovery/recovery.qdoc
```diff
@@ -203,7 +207,19 @@
...
+        -# get destination groups to which key must be recovered and call server-send for it
+        -# check server-send result for the key: if operation failed, then move key to the other bucket if possible, otherwise to 'rest-keys'
```
shaitan: Can you describe how a key migrates between different buckets including 'rest_keys' bucket.

recovery/recovery.qdoc
```diff
@@ -195,6 +195,10 @@
   -# goes throw merged results and dumps all iterated keys.
     This dump file could be used for resuming recovery after some issues.
     Dump file has name 'dump' and is placed in tmp_dir specified by <b>-D DIR, --dir=DIR</b>.
+  -# distributes keys among multiple files (buckets).
```
shaitan:
```
-# distributes keys among buckets:
    one bucket with name 'rest_keys' will contain keys which can't be recovered via server-send
    number of group buckets with name 'bucket_xx' ('xx' is group_id) will contain keys
    which can be recovered via server-send from corresponding group.
```

recovery/elliptics_recovery/types/dc.py
```diff
@@ -314,6 +371,10 @@ def lookup_keys(ctx):
     session = elliptics.Session(node)
     session.trace_id = ctx.trace_id
     filename = os.path.join(ctx.tmp_dir, 'merged_result')
+    ctx.rest_keys_filename = os.path.join(ctx.tmp_dir, 'rest_keys')
+    ctx.rest_file = open(ctx.rest_keys_filename, 'wb')
```
shaitan: `ctx.rest_keys_filename` can be get via `cx.rest_file.name`.

recovery/elliptics_recovery/types/dc.py
```diff
@@ -222,23 +227,71 @@ def contains(addresses_with_backends, address, backend_id):
...
-                os.remove(res[1])
-
-    log.debug("Merged_filename: %s, address: %s, groups: %s, tmp_dir: %s",
-              ctx.merged_filename, ctx.address, ctx.groups, ctx.tmp_dir)
+def dump_key(ctx, rest_file, key, key_infos, newest_key_group):
+    if key_infos[0].group_id != newest_key_group:
+        index = [k.group_id for k in key_infos].index(newest_key_group)
+        tmp = list(key_infos)
+        tmp[0], tmp[index] = tmp[index], tmp[0]
+        key_infos = tuple(tmp)
```
shaitan: What is the logic behind the code above?

recovery/elliptics_recovery/types/dc.py
```diff
@@ -340,11 +401,21 @@ def lookup_keys(ctx):
                         stats.counter("lookups", -1)
                 if len(key_infos) > 0:
                     key_data = (id, key_infos)
-                    dump_key_data(key_data, merged_f)
+                    if not skip_key_data(ctx, key_data):
+                        key_infos.sort(key=lambda x: (x.timestamp, x.size), reverse=True)
+                        newest_key_group = key_infos[0].group_id
+                        dump_key_data(key_data, merged_f)
+                        dump_key(ctx, ctx.rest_file, id, key_infos, newest_key_group)
```
shaitan: Why does it dump keys twice?

recovery/elliptics_recovery/types/dc.py
```diff
@@ -222,23 +227,71 @@ def contains(addresses_with_backends, address, backend_id):
...
-        with open(dump_filename, 'wb') as df:
-            for res in (r for r in results if r):
-                shutil.copyfileobj(open(res[0], 'rb'), mf)
-                os.remove(res[0])
-                shutil.copyfileobj(open(res[1], 'rb'), df)
-                os.remove(res[1])
-
-    log.debug("Merged_filename: %s, address: %s, groups: %s, tmp_dir: %s",
-              ctx.merged_filename, ctx.address, ctx.groups, ctx.tmp_dir)
+def dump_key(ctx, rest_file, key, key_infos, newest_key_group):
```
shaitan: It has `ctx`, why it doesn't use `ctx.rest_file` instead of `rest_file`?

recovery/elliptics_recovery/types/dc.py
```diff
@@ -222,23 +227,71 @@ def contains(addresses_with_backends, address, backend_id):
...
+One bucket is 'rest_keys' and other buckets are 'bucket_xx', where xx == group_id.
+'bucket_xx' contains newest keys that should be recovered from group xx to other groups
+via server_send. If a key could not be recovered with server_send it is placed to 'rest_keys'.
+
+Also this function prepares bucket_order array. The array contains group_id's sorted by amount
+of keys in appropriate bucket. This array will be used by server_send recovery process.
+'''
+def fill_buckets(ctx, results):
+    newest_key_stats = dict()
+    for _, _, range_stats in (r for r in results if r):
```
shaitan: As I see, wherever `results` are used they are filtered via `(r for r in results if r)`. It can be done ones [here] (https://github.com/reverbrain/elliptics/pull/668/files#diff-302f790c9667058c7ff408b010833a2fR328).

recovery/elliptics_recovery/dc_recovery.py
```diff
@@ -402,6 +404,12 @@ def run_one(self):
         return False
 
 
+def cleanup(ctx):
+    for f in ctx.bucket_files.itervalues():
+        os.remove(f.name)
```
shaitan: Why doesn't it close `f`?

recovery/elliptics_recovery/dc_recovery.py
```diff
@@ -412,7 +420,9 @@ def recover(ctx):
                                  net_thread_num=4,
                                  io_thread_num=24,
                                  remotes=ctx.remotes)
+    ServerSendRecovery(ctx, node).recover()
```
shaitan: Why doesn't result of `ServerSendRecovery().recover()` affect to final result of whole recovery? What if it is failed and nothing can be recovered by `WindowDC()`.

recovery/elliptics_recovery/dc_recovery.py
abudnik: ctx.rest_file used in WindowDC

agend:
Guys, do you have plan for this feature?

shaitan:
@abudnik is making changes to merge, so latter this week there will be another PR for merge recovery. After that both PRs will be reviewed and tested.

agend:
@shaitan Cool! Thanks!

abudnik:
git push forced here: this branch was forked from https://github.com/abudnik/elliptics/tree/recovery_ell - common changes in elliptics for both new dc (this PR) and coming merge (PR in process) recoveries.

bioothod:
There are comments by @shaitan which are not yet addressed. It would be great to have more comments about design of the new classes, the logic behind them and so on.
It should help understanding the whole picture of the new recovery process.

bioothod:
@abudnik there are some comments on this pull request, please resolve them and I will merge this pull request

recovery/elliptics_recovery/dc_recovery.py
abudnik: Fixed

recovery/elliptics_recovery/dc_recovery.py
abudnik: Fixed

recovery/elliptics_recovery/types/dc.py
abudnik: Fixed

recovery/elliptics_recovery/types/dc.py
abudnik: Fixed, it uses ctx.rest_file

recovery/elliptics_recovery/types/dc.py
abudnik: Fixed

recovery/elliptics_recovery/utils/misc.py
abudnik: Fixed

recovery/elliptics_recovery/types/dc.py
abudnik: Fixed. I removed rest_keys_filename from ctx.

abudnik:
@bioothod: >Does it need new tests or the old ones are enough?
It seems that old tests cover most important cases.

recovery/elliptics_recovery/types/dc.py
abudnik: First key_info in key_infos list must correspond to largest bucket among other buckets. `newest_key_group` is a group_id of the largest bucket.

recovery/recovery.qdoc
abudnik: Fixed

recovery/recovery.qdoc
abudnik: Done

recovery/recovery.qdoc
abudnik: Done

recovery/recovery.qdoc
abudnik: Fixed: `-# iterate over bucket keys by batch:`

recovery/elliptics_recovery/dc_server_send.py
```diff
@@ -0,0 +1,355 @@
...
+                    if corrupted_keys:
+                        self._remove_corrupted_keys(corrupted_keys, [group])
+
+            if timeouted_keys:
+                self._on_server_send_timeout(timeouted_keys, key_infos_map, group)
+
+    def _check_server_send_results(self, iterator, key_infos_map, group):
+        '''
+        Check result of remote sending for every key.
+        Returns list of timeouted keys.
```
bioothod: Why do you call them timeouted keys? It looks like it returns a list of failed keys no matter what was the reason

recovery/elliptics_recovery/dc_server_send.py
bioothod: This is not a documentation, this is a one-liner which says the purpose of the function without description of the logic and idea behind it. Its like saying 'this is a table' pointing your finger to table. You do not say that it is a wood table or that it is needed to host plates or books or why did you bought it and put it exactly here.

In particular, bucket/server_send classes completely lack description of the recovery process when server sending fails. This is just one note among others related to documentation. Please write not only what particular entity does, but what did you want it to do, how you want it to behave and why is this needed.

recovery/elliptics_recovery/dc_server_send.py
```diff
@@ -0,0 +1,355 @@
...
+            for k in keys:
+                result = self.remove_session.remove(k)
+                results.append(result)
+
+            timeouted_keys = []
+            timeouted_groups = set()
+            is_last_attempt = (attempt == self.ctx.attempts - 1)
+            for i, r in enumerate(results):
+                status = r.get()[0].status
+                log.info("Removing corrupted key: {0}, status: {1}, last attempt: {2}".format(keys[i], status, is_last_attempt))
```
bioothod: Do you blindly remove keys with invalid checksum even if there are no other better replicas? Should it recover these keys by reading them without checksum verification?

I do have keys which seem like broken, but they are actually correct, but for some reason they do not have checksum footer and do not have no-checksum bit in their header. I do not want them to be removed.

recovery/elliptics_recovery/types/dc.py
```diff
@@ -225,18 +230,80 @@ def final_merge(ctx, results):
...
+
+def fill_buckets(ctx, results):
+    '''
+    This function distributes keys among multiple files (buckets).
+    One bucket is 'rest_keys' and other buckets are 'bucket_xx', where xx == group_id.
+    'bucket_xx' contains newest keys that should be recovered from group xx to other groups
+    via server_send. If a key could not be recovered with server_send it is placed to 'rest_keys'.
+
+    Also this function prepares bucket_order array. The array contains group_id's sorted by amount
+    of keys in appropriate bucket. This array will be used by server_send recovery process.
```
bioothod: This is example of good documentation in this pull request.

recovery/elliptics_recovery/dc_server_send.py
```diff
@@ -0,0 +1,274 @@
...
+        if len(key_infos) > 1:
+            return key_infos[1].group_id
+        return -1
+
+    def _update_stats(self, start_time, processed_keys, recovers_in_progress, status):
+        speed = processed_keys / (time.time() - start_time)
+        recovers_in_progress -= processed_keys
+        self.stats.set_counter('recovery_speed', round(speed, 2))
+        self.stats.set_counter('recovers_in_progress', recovers_in_progress)
+        self.stats.counter('recovered_keys', 1 if status == 0 else -1)
```
bioothod: What do these new fields mean? Should the be described in qdoc?

recovery/elliptics_recovery/dc_server_send.py
abudnik: Timeouted and corrupted lists are filled in _on_server_send_fail().

recovery/elliptics_recovery/dc_server_send.py
abudnik: You can use --safe flag. From help message:
```
  -S, --safe            Do not remove recovered keys after merge [default:
                        False]
```

recovery/elliptics_recovery/dc_server_send.py
abudnik: See docs in recovery/recovery.qdoc ('Server-send recovery: ...')

bioothod: This doesn't tell me how `on_server_send_fail()` behaves in bucket manager and server-send recovery classes and how they are related to each other.

Documentation in the code tells reader how particular **code** is supposed to be designed and how it is supposed to be working. Text is qdoc you reference above tells higher-level user how recovery works.

recovery/elliptics_recovery/dc_server_send.py
bioothod: What about recovering such keys?
With this option, keys will not be read, but they also will not be recovered, is that correct?

recovery/elliptics_recovery/dc_server_send.py
bioothod: There is also `buckets.on_server_send_fail()` call, which does something with the key. Will it change the key? Or put into some other place?

Please describe how these new classes should work, what is the design decision behind this code, it is vital knowledge to understand and fix bugs.

recovery/elliptics_recovery/dc_server_send.py
abudnik: With this option, keys will not be removed.
Keys with invalid checksum will not be recovered via server-send, because they have invalid checksum.

recovery/elliptics_recovery/dc_server_send.py
abudnik: Added more comments on handling server-send (recovery) failures: https://github.com/abudnik/elliptics/commit/cbbad2448b5a2da2d7e86b41b364f38edb51ae1d

abudnik:
This PR is ready to be merged.

PR #670, opened 2015-11-24
tests: separate test targets for running test modules individually

ijon:
New `add_test_target` function is used to define a test module: its make target and any dependencies that are required to run it.

This creates the ability to build and run test modules individually (which is useful in development).

`make test` builds and runs all test modules in one go, same as before.

But, for example, `make test_srw` or `make test_api` will only build and run `dnet_cpp_srw_test` or `dnet_cpp_api_test` module.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -42,65 +42,149 @@ if(WITH_COCAINE)
 endif()
 target_link_libraries(test_common ${TEST_COMMON_LIBRARIES})
 
-add_executable(dnet_cpp_test test.cpp)
-set_target_properties(dnet_cpp_test ${TEST_PROPERTIES})
-target_link_libraries(dnet_cpp_test ${TEST_LIBRARIES})
+# add_test_target({target} {test-executable} DEPENDS {required-dependencies})
+#
+# Defines target for running a test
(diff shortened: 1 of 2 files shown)


tests/CMakeLists.txt
```diff
@@ -42,65 +42,149 @@ if(WITH_COCAINE)
...
-add_executable(dnet_cpp_test test.cpp)
-set_target_properties(dnet_cpp_test ${TEST_PROPERTIES})
-target_link_libraries(dnet_cpp_test ${TEST_LIBRARIES})
+# add_test_target({target} {test-executable} DEPENDS {required-dependencies})
+#
+# Defines target for running a test
+# Test executable is a target for building a test.
+# Dependencies are not only build dependencies but also runtime dependencies.
+# (Some tests, for example, launch dnet_ioserv directly,
+# so it must be build for those tests to be able to run)
```
bioothod: Please describe in comments what should be done by test author to enable new set of tests.
I.e. what cmake function he has to run/register with, which list it should add new file into and so on.

ijon: Ok, I'll add some description.

ijon:
Added more comments and instruction on how to add new C++ test module.

PR #671, opened 2015-11-24
Fix cmake testenv handling

ijon:
MERGE AFTER #670.

`add_custom_target` does not directly support setting environment for the given COMMAND, thus forcing everybody to use `A=a B=b cmd` shell syntax. But in some cases it does not work properly, as `add_custom_target` erroneously adds additional escaping and result command becomes incorrect.

Adding explicit quotes in proper places and using `env` command helps.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -42,65 +42,149 @@ if(WITH_COCAINE)
 endif()
 target_link_libraries(test_common ${TEST_COMMON_LIBRARIES})
 
-add_executable(dnet_cpp_test test.cpp)
-set_target_properties(dnet_cpp_test ${TEST_PROPERTIES})
-target_link_libraries(dnet_cpp_test ${TEST_LIBRARIES})
+# add_test_target({target} {test-executable} DEPENDS {required-dependencies})
+#
+# Defines target for running a test
+# Test executable is a target for building a test.
+# Dependencies are not only build dependencies but also runtime dependencies.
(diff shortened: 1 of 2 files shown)


tests/CMakeLists.txt
```diff
@@ -42,65 +42,149 @@ if(WITH_COCAINE)
...
-add_executable(dnet_cpp_test test.cpp)
-set_target_properties(dnet_cpp_test ${TEST_PROPERTIES})
-target_link_libraries(dnet_cpp_test ${TEST_LIBRARIES})
+# add_test_target({target} {test-executable} DEPENDS {required-dependencies})
+#
+# Defines target for running a test
+# Test executable is a target for building a test.
+# Dependencies are not only build dependencies but also runtime dependencies.
+# (Some tests, for example, launch dnet_ioserv directly,
+# so it must be build for those tests to be able to run)
```
bioothod: These changes are already present in pull request https://github.com/reverbrain/elliptics/pull/670, aren't they?

ijon: #670 and #671 should be applied in order, #671 works on top of #670.
As pull requests are actually just a diffs between pr-branch and master branch, then after #670 will be merged into master, content of this, #671, PR will be the single commit 528cc52. As intended.

ijon:
Had to do push -f. Still https://github.com/reverbrain/elliptics/pull/671#discussion_r45979740 applies: merge #670 first, this one after that.

PR #672, opened 2015-11-24
Core: bypass write to pipe for local transactions

toshic:
Reworked dnet_io_req_queue function to pass io_req struct directly
to the proper IO pool, eliminating write/read through pipe in net pool.

### DIFF
--- a/library/net.c
+++ b/library/net.c
@@ -171,14 +171,28 @@ void dnet_state_clean(struct dnet_net_state *st)
 	dnet_log(st->n, DNET_LOG_NOTICE, "Cleaned state %s, transactions freed: %d", dnet_state_dump_addr(st), num);
 }
 
-static struct dnet_io_req *dnet_io_req_copy(struct dnet_net_state *st, struct dnet_io_req *orig)
+/*
+ * This function makes a copy of io_req to transfer memory ownership to another thread.
+ * If target thread is net thread we can just make a copy and propertly fill fd params.
+ * If target thread is io thread (bypass for local transaction on server side, read_data flag
+ * is set to 1) we need to allocate buffer and read fd content info this buffer.
+ * Result should looks exactly as if was read from network socket. This CPU IO time is spent
+ * in backend's IO pool.
+ */
+static struct dnet_io_req *dnet_io_req_copy(struct dnet_net_state *st, struct dnet_io_req *orig, int read_data)
 {
 	void *buf;
+	size_t len = 0;
 	struct dnet_io_req *r;
 	int offset = 0;
 	int err = 0;
 
-	buf = r = malloc(sizeof(struct dnet_io_req) + orig->dsize + orig->hsize);
+	len = sizeof(struct dnet_io_req) + orig->dsize + orig->hsize;
+	if (orig->fd >= 0 && orig->fsize && read_data) {
+		len += orig->fsize;
+	}
+
+	buf = r = malloc(len);
 	if (!r) {
 		dnet_log(st->n, DNET_LOG_ERROR, "Not enough memory for io req queue fd: %d : %s %d", orig->fd, strerror(-err), err);
 		return NULL;
@@ -192,7 +206,10 @@ static struct dnet_io_req *dnet_io_req_copy(struct dnet_net_state *st, struct dn
 
 		offset = r->hsize;
 		memcpy(r->header, orig->header, r->hsize);
-	}
+	} else if (read_data) {
+        r->header = buf + sizeof(struct dnet_io_req);
+        r->hsize = 0;
+    }
 
 	if (orig->data && orig->dsize) {
 		r->data = buf + sizeof(struct dnet_io_req) + offset;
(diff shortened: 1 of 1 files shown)


library/net.c
```diff
@@ -192,7 +206,10 @@ static struct dnet_io_req *dnet_io_req_copy(struct dnet_net_state *st, struct dn
 
 		offset = r->hsize;
 		memcpy(r->header, orig->header, r->hsize);
-	}
+	} else if (read_data) {
+        r->header = buf + sizeof(struct dnet_io_req);
```
shaitan: If `orig` doesn't have header, why does it set `r->header` here?

library/net.c
```diff
@@ -171,14 +171,28 @@ void dnet_state_clean(struct dnet_net_state *st)
 	dnet_log(st->n, DNET_LOG_NOTICE, "Cleaned state %s, transactions freed: %d", dnet_state_dump_addr(st), num);
 }
 
-static struct dnet_io_req *dnet_io_req_copy(struct dnet_net_state *st, struct dnet_io_req *orig)
+/*
+ * This function makes a copy of io_req to transfer memory ownership to another thread.
+ * If target thread is net thread we can just make a copy and propertly fill fd params.
```
shaitan: `propertly` -> `properly`

library/net.c
```diff
@@ -221,7 +258,41 @@ static int dnet_io_req_queue(struct dnet_net_state *st, struct dnet_io_req *orig
 	int err = 0;
 	struct dnet_io_req *r;
 
-	r = dnet_io_req_copy(st, orig);
+	/*
+	 * If destination is local node there is no need to copy it via net thread.
+	 * Here it could be sheduled to the proper IO pool directly.
```
shaitan: `sheduled` -> `scheduled`

library/net.c
```diff
@@ -221,7 +258,41 @@ static int dnet_io_req_queue(struct dnet_net_state *st, struct dnet_io_req *orig
...
+
+		/*
+		 * Fixup request.
+		 * r->header and r->data are just a pointers to the same buffer.
+		 * Here r->header could be any size, it depend on function that created io_req.
+		 * In net thread r->hsize is ALWAYS sizeof(struct dnet_cmd) and r->data points
+		 * exactly after struct dnet_cmd struct. We need to modify our copy to the same layout
+		 * because dnet_schedule_io requires it.
+		 */
+		if (r->hsize != sizeof(struct dnet_cmd)) {
```
shaitan: Why does this magic live outside of `dnet_io_req_copy`?

bioothod: Actually strict contiguous layout for header and data is not required, it should be ok to put them into different buffers with different sizes, but this would need to extend `dnet_io_req_free()`

library/net.c
```diff
@@ -336,8 +407,12 @@ ssize_t dnet_send(struct dnet_net_state *st, void *data, uint64_t size)
 	struct dnet_io_req r;
 
 	memset(&r, 0, sizeof(r));
-	r.data = data;
-	r.dsize = size;
+	r.header = data;
+	r.hsize = sizeof(struct dnet_cmd);
+	if (size > r.hsize) {
```
shaitan: It should return `-EINVAL` if `size` is smaller than `sizeof(struct dnet_cmd)`.

bioothod:
Why is this needed? Do you have so many local requests which come from one queue to another?
What are these commands, and why do they need this special handling?

toshic:
We want to run traffic-intensive wokers with SRW, and use Cocaine localnode service (it will be in separate PR later) to access data. We do not want to start elliptics client in each worker because our cluster is pretty large, and we can't update it (elliptics client inside worker's code) in case of Elliptics cluster major version upgrade.

shaitan:
@toshic It is ready to be merged, isn't it?

toshic:
I shink so.

library/net.c
bioothod: That's invalid comment.

Header size must be at least sizeof(struct dnet_cmd) when reading valid elliptics command, how is it possible that r->header can be of any size?
Also, code below breaks if header size + data size are less than sizeof(struct dnet_cmd).

Why is this needed at all? Why would caller fail to set pointers and sizes correctly?

toshic: For dnet_schedule_io header size must be sizeof(struct dnet_cmd). But when io_req is creating inside elliptics code header size could be more than sizeof(struct dnet_cmd) and header field could have dnet_io_attr inside for example. Here we need to fix this, dnet_io_attr struct position inside buffer should be after io_req->data pointer.

bioothod: >For dnet_schedule_io header size must be sizeof(struct dnet_cmd). 

That's wrong assumption, header must be at least `sizeof(struct dnet_cmd)` since this is how elliptics protocol works.
Header size is a boundary which is needed when **sender** wants to send some preallocated data plus data from file descriptor, for instance sending read header (`struct dnet_cmd` + `struct dnet_io_attr`) + eblob content from file descriptor via `sendfile()`.

Receiver side only uses header size and pointer for informational purposes since receiver reads whole command content (all headers + data) into single continuous buffer.

toshic: That how elliptics works right now. I've tried to copy request as is, and dnet_schedule_io fails. This fixed header size if guarantied by socket reading function, in dnet_process_recv_single.

bioothod: How did `dnet_schedule_io` fail? It doesn't use `r->hsize` field at all, but expects `r->header` to point to valid `struct dnet_cmd`, and later processing functions expect `r->data` to be continuous.

toshic: dnet_schedule_io assumes that r->header field consist only dnet_cmd and everythind else is inside data. This structure is provided by dnet_process_recv_single.
After it selects correct IO pool it passes r->data into it. If dnet_io_attr is inside r->header, r->data will be NULL and we'll get SIGSEGV.

bioothod: I do not see where `dnet_schedule_io` assumes that header is exactly `sizeof(dnet_cmd)` bytes, that would be strange.
It requires pointers to be set correctly, even more, `r->header` pointer at receive side is only needed to print some debug messages. There is no need to manipulate header sizes at all. But they should be set for correctness.

Please drop this check and setup sender code correctly - set `r->header` and `r->data` to point to the same buffer with whole command: `dnet_cmd` + `dnet_exec` + data, and set `r->hsize` to be equal to at least `sizeof struct dnet_cmd`, `r->st` should grab the reference to the receiving state, `r->fd` should be -1.

PR #675, opened 2015-11-26
Server send dynamic queue

bioothod:
Implemented dynamic queue size for copy/move iterator.
Instead of hardcoded (30G previously) value maximum number of bytes written and not yet acknowledged is calculated dynamically based on time every write request took to complete.
Queue size equals to 5 seconds of data written into the wire at the speed of the just finished write request. 5 seconds has been selected as it equals to default operation timeout.

This pull requests passes all tests, but also passes real-life multi-terabyte tests, which previously failed to complete with slow 'destination' disks.

### DIFF
--- a/library/dnet.c
+++ b/library/dnet.c
@@ -319,7 +319,7 @@ static void dnet_queue_wait_threshold(struct dnet_net_state *st)
 	/* If send succeeded then we should increase queue size */
 	while ((atomic_read(&st->send_queue_size) > DNET_SEND_WATERMARK_HIGH) && !st->__need_exit) {
 		/* If high watermark is reached we should sleep */
-		dnet_log(st->n, DNET_LOG_DEBUG,
+		dnet_log(st->n, DNET_LOG_ERROR,
 				"State high_watermark reached: %s: %ld, sleeping",
 				dnet_addr_string(&st->addr),
 				atomic_read(&st->send_queue_size));
@@ -331,7 +331,7 @@ static void dnet_queue_wait_threshold(struct dnet_net_state *st)
 			pthread_cond_wait(&st->send_wait, &st->send_lock);
 		pthread_mutex_unlock(&st->send_lock);
 
-		dnet_log(st->n, DNET_LOG_DEBUG, "State woken up: %s: %ld",
+		dnet_log(st->n, DNET_LOG_ERROR, "State woken up: %s: %ld",
 				dnet_addr_string(&st->addr),
 				atomic_read(&st->send_queue_size));
 	}
@@ -517,7 +517,7 @@ err_out_send:
 			if (err && !send->write_error)
 				send->write_error = err;
 
-			if (atomic_sub(&send->bytes_pending, resize) < DNET_SERVER_SEND_WATERMARK_LOW)
+			if ((atomic_sub(&send->bytes_pending, resize) < DNET_SERVER_SEND_WATERMARK_LOW) || send->write_error)
 				pthread_cond_broadcast(&send->write_wait);
 
 			dnet_server_send_put(send);
--- a/library/dnet.c
+++ b/library/dnet.c
@@ -517,7 +517,7 @@ err_out_send:
 				new_pending = 5 * re->size * 1000000 / usec;
 			}
 
-			dnet_log(st->n, DNET_LOG_NOTICE, "%s: %s: sending response to client: %s, "
+			dnet_log(st->n, DNET_LOG_INFO, "%s: %s: sending response to client: %s, "
 					"user_flags: %llx, ts: %s (%lld.%09lld), "
 					"status: %d, size: %lld, iterated_keys: %lld/%lld, write_error: %d, "
 					"pending_bytes: %ld, limit: %ld -> %ld",
@@ -760,7 +760,7 @@ int dnet_server_send_write(struct dnet_server_send_ctl *send,
 
 	atomic_add(&send->bytes_pending, re->size);
 
-	dnet_log(n, DNET_LOG_NOTICE, "%s: %s: sending WRITE request, iterator response: %s, user_flags: %llx, ts: %s (%lld.%09lld), "
+	dnet_log(n, DNET_LOG_INFO, "%s: %s: sending WRITE request, iterator response: %s, user_flags: %llx, ts: %s (%lld.%09lld), "
 			"status: %d, size: %lld, iterated_keys: %lld/%lld",
 			__func__,
 			dnet_dump_id(&send->cmd.id), dnet_dump_id_str(re->key.id),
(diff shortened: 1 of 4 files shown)


library/elliptics.h
```diff
@@ -1035,8 +1035,7 @@ static inline int dnet_read_ll(int fd, char *data, size_t size, off_t offset)
 /*
  * Watermarks for number of bytes written into the wire
  */
-#define DNET_SERVER_SEND_WATERMARK_HIGH		(30*1024*1024*1024L)
-#define DNET_SERVER_SEND_WATERMARK_LOW		DNET_SERVER_SEND_WATERMARK_HIGH / 2
+#define DNET_SERVER_SEND_WATERMARK_HIGH		(100*1024*1024L)
```
abudnik: Does it mean that we cannot fully utilize 10Gb networks with fast raid/ssd destination disks?

abudnik:
Most interesting performance test (imho) is sending sequence of keys with monotonically decreasing data sizes / with combination of small and big keys.

abudnik:
Also this approach dynamically changes queue size on periodical slowdowns of a disk (e.g. blob defragmentation, periodical disk flushes, ...), while average disk speed may be fine.

library/elliptics.h
bioothod: It is unlikely that you want to fully utilize any network in the first place, since you also need a 'space' in the wire for client requests and optionally other backend recovery.

But nevertheless 100 megabytes of pending unacknowledged data is quite large chunk and it is calculated after key has been submitted into the wire, i.e. iterator sends current key and then sleeps if number of unacknowledged bytes is higher than this limit. If you have fast network and remote storage and big keys, this limit can be exceeded.

bioothod:
I ran performance test with production data and got sustained 80MB/s for overwrite mode and 50MB/s for cas-timestamp mode. There are keys from several hundreds of bytes to gigabyte.

It indeed changes queue size for rather small slowdowns, and that's the purpose.
Queue size is quickly restored as soon as write completes quickly.

library/elliptics.h
abudnik: JFYI, new server-send recovery speed is bounded by network bandwidth (120 mb/sec) for 128kb+-10kb keys and those numbers are taken from current performance test results of new s.s. recovery.

bioothod: Good to know. What was the total data size and amount of RAM on accepting node?

abudnik: total data size 100 Gb
RAM 128 Gb

bioothod: That's not a good dataset, it fits ram, try increasing data size by order of magnitude.
Existing server-send will likely timeout if remote disks will not cope with the rate, newer server-send will converge to disk speed and will float around when flush/sync happens.

library/dnet.c
```diff
@@ -319,7 +319,7 @@ static void dnet_queue_wait_threshold(struct dnet_net_state *st)
 	/* If send succeeded then we should increase queue size */
 	while ((atomic_read(&st->send_queue_size) > DNET_SEND_WATERMARK_HIGH) && !st->__need_exit) {
 		/* If high watermark is reached we should sleep */
-		dnet_log(st->n, DNET_LOG_DEBUG,
+		dnet_log(st->n, DNET_LOG_ERROR,
```
shaitan: Why have you changed log level?

bioothod: Because this is an error in the first place, and also this may help debugging the case, when elliptics suddenly eats all available ram without sending any single reply to clients.

shaitan: Why is it an error? Yes, we have reached the high watermark and should sleep, but it is common situation, isn't it?

bioothod: This means that client is not capable of getting server replies at the rate it has requested. There might be a lot of reasons, why this has happened, but it is an error - requested action can not be done.

Also, this puts IO thread into sleep, which in turn means other requests will not be processed, which is another reason why this is an error.

shaitan: I still don't see an error here, so I prefer using NOTICE/INFO level because on fast device it will spam logs without any help. It is better to add logs around similar threshold in server-send.

bioothod: There is info message for every write request in server-send logic, which includes size of the queue and its limit.

I've decreased queue size log level to notice.

shaitan:
@bioothod, can you provide detailed results of your performance tests for this PR?

library/dnet.c
```diff
@@ -499,9 +501,26 @@ static int dnet_iterator_server_send_complete(struct dnet_addr *addr, struct dne
...
+				gettimeofday(&tv, NULL);
+
+				usec = (tv.tv_sec - wp->start.tv_sec) * 1000000 + (tv.tv_usec - wp->start.tv_usec);
+
+				/*
+				 * Maximum number of bytes written into the wire and not yet acknowledged.
+				 * It is equal to 'current speed' times 5 seconds, since 5 seconds is default wait timeout
+				 * for write command.
+				 */
+				new_pending = 5 * re->size * 1000000 / usec;
```
shaitan: Why does it use 5 seconds timeout here, but for write it use 60 seconds timeout?

library/elliptics.h
shaitan: How `DNET_SERVER_SEND_WATERMARK_HIGH` relates to `'current speed' times 5 seconds`?

library/dnet.c
```diff
@@ -717,6 +749,8 @@ int dnet_server_send_write(struct dnet_server_send_ctl *send,
 
 	dnet_session_set_trace_id(s, send->cmd.trace_id);
 	dnet_session_set_trace_bit(s, !!(send->cmd.flags & DNET_FLAGS_TRACE_BIT));
+	if (dnet_session_get_timeout(s)->tv_sec < 60)
+		dnet_session_set_timeout(s, 60);
```
shaitan: What was the purpose for this?

library/dnet.c
bioothod: 5 seconds is our default timeout, I used this number for upper boundary, this is a heuristic number, which allows iterator to actually copy data over 1gbit network between common disks without timeouts, in theory it might be higher, but there is a risk of overfilling the disk bandwidth.

library/dnet.c
bioothod: There are pikes in my environment which stop destination backend from accepting new requests for more than several seconds, so I increased default lower bound for write timeout.

library/elliptics.h
bioothod: It doesn't, the former is higher bound of the tube, the latter is dynamic algorithm which takes into account 5 seconds of data.

bioothod:
Performance test for dynamically sized queue on 2Tbs of production data ranging from several hundreds of bytes to gigabyte:
* sustained 80MB/s for overwrite mode both when destination backend contains data and when it doesn't
* 50MB/s for cas-timestamp mode

In both cases this is a disk limit and not software bounds.
Without dynamically sized queue iterator filled destination node with 120MB/s speed and started to timeout after a while, because disk was not able to write data at the pace it has been receiving incoming packets, they stayed in queue for too long and eventually client timed out.

I used `example/dnet_iterate_move` tool to copy/move data.

library/dnet.c
shaitan: So, it is supposed that 60 seconds timeout should be good enough for everyone. Let's make 60 as a constant with comments about pikes reason.

bioothod: I've added comments

PR #676, opened 2015-12-02
binding: python: added server_send & start_copy_iterator session methods

abudnik:

PR #678, opened 2015-12-07
merge recovery: optimization: use copy-iterator & server-send for small/medium keys

abudnik:


recovery/elliptics_recovery/iterator.py
```diff
@@ -379,6 +386,19 @@ def iterate_with_stats(cls, node, eid, timestamp_range,
         return result, result_len
 
 
+class CopyIterator(Iterator):
+    def __init__(self, *args,  **kwargs):
+        super(CopyIterator, self).__init__(*args, **kwargs)
+
+    def _start_iterator(self, eid, ranges, itype, flags, timestamp_range):
+        flags |= elliptics.iterator_flags.move
+        return self.session.start_copy_iterator(eid, ranges, [eid.group_id], flags, timestamp_range[0], timestamp_range[1])
```
bioothod: What are the destination groups?

bioothod:
This pull request contains fair amount of code from #676, presumably it depends on #676, but the former is not yet ready, is that correct?

recovery/elliptics_recovery/types/merge.py
```diff
@@ -686,6 +683,98 @@ def succeeded(self):
         return self.result
 
 
+class ServerSendRecovery(object):
```
bioothod: Please add description of how does this class work, what is the logic behind it and so on, as much information as possible

recovery/elliptics_recovery/iterator.py
abudnik: CopyIterator is used in merge recovery only, that's why exactly one group specified here.

bioothod: `CopyIterator` name implies the opposite, please add documentation describing how it is used and better change the name

abudnik:
>This pull request contains fair amount of code from #676, presumably it depends on #676, but the former is not yet ready, is that correct?

Yes. Also, #668 depends on #676.

recovery/elliptics_recovery/iterator.py
abudnik: Ok

abudnik: https://github.com/abudnik/elliptics/commit/f7bd8c5c29e233d71b0a3b788c1e5e3643023353

recovery/elliptics_recovery/types/merge.py
abudnik: done

recovery/elliptics_recovery/types/dc.py
```diff
@@ -69,9 +69,7 @@ def iterate_node(arg):
             batch_size=ctx.batch_size,
             stats=stats,
             flags=flags,
-            leave_file=True,
-            separately=True,
-            trace_id=ctx.trace_id)
```
shaitan: Why did you remove `trace_id`?

recovery/elliptics_recovery/iterator.py
```diff
@@ -379,6 +386,24 @@ def iterate_with_stats(cls, node, eid, timestamp_range,
...
+    '''
+    def __init__(self, *args,  **kwargs):
+        super(MergeRecoveryIterator, self).__init__(*args, **kwargs)
+
+    def _start_iterator(self, eid, ranges, flags, timestamp_range):
+        flags |= elliptics.iterator_flags.move
+        return self.session.start_copy_iterator(eid, ranges, [eid.group_id], flags, timestamp_range[0], timestamp_range[1])
+
+    def _on_key_response(self, results, record):
+        if record.response.status != 0:
```
shaitan: It should also update statistics in own way, because now it will not show any recovering progress and all iterator's statistics will be about keys which weren't recovered by copy-iterator.

recovery/elliptics_recovery/types/merge.py
```diff
@@ -686,6 +683,130 @@ def succeeded(self):
...
+        self.routes = self._prepare_routes(ctx, group)
+        self.session = elliptics.Session(node)
+        self.session.exceptions_policy = elliptics.exceptions_policy.no_exceptions
+        self.session.set_filter(elliptics.filters.all)
+        self.session.timeout = 60
+        self.session.groups = [group]
+        self.session.trace_id = ctx.trace_id
+        self.ctx = ctx
+
+    def _prepare_routes(self, ctx, group):
```
shaitan: Is it really needed? Why not just use `routes.get_id_routes()` for filtering keys?

recovery/elliptics_recovery/types/merge.py
```diff
@@ -702,12 +823,15 @@ def dump_process_group((ctx, group)):
...
     ret = True
     with open(ctx.dump_file, 'r') as dump:
+        ss_rec = ServerSendRecovery(ctx, node, group)
         # splits ids from dump file in batchs and recovers it
         for batch_id, batch in groupby(enumerate(dump), key=lambda x: x[0] / ctx.batch_size):
             recovers = []
             rs = RecoverStat()
-            for _, val in batch:
-                rec = DumpRecover(routes=ctx.routes, node=node, id=elliptics.Id(val), group=group, ctx=ctx)
+            keys = [elliptics.Id(val) for _, val in batch]
```
shaitan: Use generator here to avoid unnecessary iteration over keys.

recovery/elliptics_recovery/types/merge.py
```diff
@@ -686,6 +683,130 @@ def succeeded(self):
...
+        removes keys with older timestamp or invalid checksum.
+        Returns list of keys that was not recovered via server-send.
+        '''
+        log.info("Server-send bucket: num keys: {0}".format(len(keys)))
+
+        def contain(key, ranges):
+            index = bisect(ranges, key)
+            return index % 2 == 1
+
+        responses = dict([(str(k), []) for k in keys]) # key -> [list of responses]
```
shaitan: It is better to use follow form to avoid unnecessary iteration and creating of temporary list:
```python
responses = {str(k): [] for k in keys}
```

recovery/elliptics_recovery/types/merge.py
```diff
@@ -686,6 +683,130 @@ def succeeded(self):
...
+        self.session.trace_id = ctx.trace_id
+        self.ctx = ctx
+
+    def _prepare_routes(self, ctx, group):
+        '''
+        Returns list of triplets (address, backend, [ranges]),
+        where ranges are sorted by their left boundary.
+        '''
+        def sort_ranges(ranges):
+            ranges = sorted(ranges, key=lambda r: r[0])
```
shaitan: `ranges` are already sorted, aren't they?

recovery/elliptics_recovery/types/merge.py
```diff
@@ -686,6 +683,130 @@ def succeeded(self):
...
+        '''
+        bad_keys = []
+        for val in responses.itervalues():
+            bad_keys.extend([r for r in val if self._check_bad_key(r)])
+
+        results = []
+        for k in bad_keys:
+            key, _, addr, backend_id = k
+            self.session.set_direct_id(addr, backend_id)
+            result = self.session.remove(key)
```
shaitan: Recovery has a safe mode (`-S/--safe`) when it doesn't remove keys. Please check that it isn't set before removing keys and before setting `elliptics.iterator_flags.move` flag.

recovery/elliptics_recovery/types/merge.py
shaitan: This class doesn't update any statistics, so it is impossible to track recovery progress.

recovery/elliptics_recovery/types/merge.py
```diff
@@ -686,6 +683,130 @@ def succeeded(self):
...
+        log.debug("Server-send: address: {0}, backend: {1}, num keys: {2}".format(addr, backend_id, len(keys)))
+
+        self.session.set_direct_id(addr, backend_id)
+        iterator = self.session.server_send(keys, elliptics.iterator_flags.move, list(self.session.groups))
+        for result in iterator:
+            status = result.response.status
+            key = result.response.key
+            r = (key, status, addr, backend_id)
+            log.debug("Server-send result: key: {0}, status: {1}".format(str(key), status))
+            responses[str(key)].append(r)
```
shaitan: Why does it need 2 copies of the key in key and value of the dict?

recovery/elliptics_recovery/types/merge.py
```diff
@@ -702,12 +823,15 @@ def dump_process_group((ctx, group)):
...
+        ss_rec = ServerSendRecovery(ctx, node, group)
         # splits ids from dump file in batchs and recovers it
         for batch_id, batch in groupby(enumerate(dump), key=lambda x: x[0] / ctx.batch_size):
             recovers = []
             rs = RecoverStat()
-            for _, val in batch:
-                rec = DumpRecover(routes=ctx.routes, node=node, id=elliptics.Id(val), group=group, ctx=ctx)
+            keys = [elliptics.Id(val) for _, val in batch]
+            keys = ss_rec.recover(keys)
+            for k in keys:
```
shaitan: Do we know cases when a key can be failed to be recovered via server-send but can be succeeded to be recovered via read/write? Can we get rid of read/write recovery here and everywhere?

bioothod: Server-send recovery will not work if servers in different groups are not connected, in particular, this happens when storage nodes are being firewalls in different physical locations and they are not connected via private network.

shaitan: Do you have such cluster or do you know who has?

bioothod: I have this cluster setup

recovery/elliptics_recovery/types/merge.py
noxiouz: If I understand properly, `ss_rec.recover` iterates more than once through `keys`, so it can not be a generator.

recovery/elliptics_recovery/types/merge.py
```diff
@@ -686,6 +683,130 @@ def succeeded(self):
...
+    def _check_bad_key(self, response):
+        status = response[1]
+        return status == -errno.EBADFD or status == -errno.EILSEQ
+
+    def _get_unrecovered_keys(self, responses):
+        '''
+        Returns keys that was not recovered via server-send.
+        '''
+        keys = []
+        for val in responses.iteritems():
```
noxiouz: ```for key_id, key_responses in responses.iteritems()``` to get rid of `val[1]` and `val[0]`

shaitan: Please, take a look at #679 where this PR was complemented.

PR #680, opened 2015-12-21
added wait-ts (timeout) to alloc/send log

shaitan:


bioothod:
Can we put debug message generation from `dnet_trans_destroy()` into common place and use it in allocation routing?

shaitan:
Yes, we can and I will make it later today.

PR #679, opened 2015-12-15
Recovery merge

abudnik:
recovery: merge: removed obsolete Recovery code; used server-send, copy-iter everywhere

This PR sends greetings to PR #678

PR #681, opened 2015-12-22
Copy iterator interrupt

abudnik:
1. copy-iterator: stop iteration only if connection failed or no space left on remote backend
2. server-send: check that source and destination backends are not the same
3. server-send: added 80 reserved bytes to dnet_server_send_request struct


abudnik:
PRs https://github.com/reverbrain/elliptics/pull/668 & https://github.com/reverbrain/elliptics/pull/679 are depend on this PR

library/dnet.c
```diff
@@ -783,6 +785,25 @@ int dnet_server_send_write(struct dnet_server_send_ctl *send,
 			(unsigned long long)re->iterated_keys, (unsigned long long)re->total_keys,
 			atomic_read(&send->bytes_pending), send->bytes_pending_max);
 
+	// check that source and destination backends are not the same
```
bioothod: Shouldn't it be on the client side?
Do not even try to start iterator if its source/destination backends are the same?

library/dnet.c
```diff
@@ -433,12 +433,13 @@ static int dnet_iterator_server_send_complete(struct dnet_addr *addr, struct dne
...
-		 * Write CAS error is not a real 'error' in that regard, that we do not remove
-		 * local record, but also do not stop iterator.
-		 *
 		 * Setting @write_error forces iterator to stop.
+		 *
+		 * Stop iteration only if connection was failed on write
+		 * or no space left on remote backend.
 		 */
-		if (err && !send->write_error && (err != -EBADFD))
+		const int critical_error = err && (err == -ENXIO || err == -ENOSPC);
```
bioothod: No, this is wrong.
Server can answer with many errors which we do not know about, we can not proceed with writing data in this case.

EBADFD is special, since it says write has failed because of CAS/timestamp-cas error, which is OK for this iterator.
Actually, we should proceed only if there is such error AND this is the timestamp-cas case.

What errors did you see that you still want to proceed?

include/elliptics/packet.h
```diff
@@ -1244,6 +1244,7 @@ struct dnet_server_send_request {
 	int		id_num;
 	int		group_num;
 	uint64_t	iflags;
+	uint64_t	reserved[10];
```
bioothod: Isn't it too much? For what do you need all that space?

bioothod:
How do https://github.com/reverbrain/elliptics/pull/668 and https://github.com/reverbrain/elliptics/pull/679 depend on this pull request?

library/dnet.c
abudnik: e.g. ENOENT, ETIMEDOUT

library/dnet.c
abudnik: No, this happens when client state gone during iteration. It starts to write to itself.

library/dnet.c
bioothod: I believe, timeout has to stop iterator - receiving side does not cope with the data rate.
I do not know when write can return ENOENT

library/dnet.c
bioothod: Ok, I see. Please add extended comments about this case

abudnik: ok

include/elliptics/packet.h
abudnik: cast @shaitan

library/dnet.c
abudnik: ENOENT happens on writing zero-size key

bioothod: Ok, we could skip it too.

I do know that for example write returns EINVAL in many cases and this should stop iterator.
Number of errors to skip is much smaller than that of to proceed, let's write them all in this place and add comments about cases when these errors happen, this will help with understanding what to expect from write.

include/elliptics/packet.h
shaitan: It is needed for future use. I suppose that 80 isn't too much, so I suggested it. Do you have another suggestion?

library/dnet.c
```diff
@@ -783,6 +785,25 @@ int dnet_server_send_write(struct dnet_server_send_ctl *send,
...
 
+	// check that source and destination backends are not the same
+	const int group_id = dnet_group_id_search_by_backend(n->st, send->backend_id);
+	if (group_id >= 0) {
+		for (i = 0; i < send->group_num; ++i) {
+			if (send->groups[i] == group_id) {
+				ctl.id.group_id = group_id;
+				const struct dnet_net_state *st = dnet_state_get_first_with_backend(n, &ctl.id, &backend_id);
+				if (st == n->st && backend_id == send->backend_id) {
+					dnet_log(n, DNET_LOG_ERROR, "%s: Interrupting iterator: source and destination backends are the same: "
```
shaitan: Does it really interrupt iterator in this case or it just skips the key?

abudnik: It interrupts iterrator. I've manually tested this code with iptables before sending PR :)

shaitan: How can it be tested via iptables? It's about writing keys when destination and source are the same backend.

abudnik: It interrupts iterrator.

shaitan: What?

abudnik: Iterator's callback returns error code from dnet_server_send_write(), so iteration stops on this error.

abudnik: I've tested manually with iptables closing remove connection, so this case was reproduced and iteration stopped.

shaitan: It shouldn't interrupt iteration here and there are few reasons for that:
* it should avoid writing keys to the same backend, but it should write all keys to other groups
* if there is only one destination group, it should skip keys that should be written to the same backend, but it should write all keys to other backends

abudnik: I agree and will fix it soon.

abudnik: Fixed: https://github.com/abudnik/elliptics/commit/ae161ba41243c3c73fd8ca406da8a50e450096ef

library/dnet.c
```diff
@@ -671,6 +672,28 @@ static int dnet_server_send_sync(struct dnet_server_send_ctl *ctl)
 	return dnet_server_send_put(ctl);
 }
 
+static int dnet_filter_invalid_groups(int *groups, const struct dnet_server_send_ctl *send, struct dnet_node *n, struct dnet_id *id)
```
bioothod: This requires more comments about the whole logic behind this function.
Why is it needed, what does it do, how does it modify group array.

One should understand this function without looking into the code, but from description.
Looking into the code is only needed when you want to fix a bug.

It also leaks state.

abudnik: Fixed: https://github.com/abudnik/elliptics/commit/a93cc635824385a52395bb1db5a2dc2d1bad6bd6

PR #682, opened 2015-12-28
session: added new server_send() method which works with vector of elliptics::key

bioothod:
This is the most generic method, others use it internally.

### DIFF
--- a/bindings/cpp/session.cpp
+++ b/bindings/cpp/session.cpp
@@ -2353,7 +2353,7 @@ async_iterator_result session::cancel_iterator(const key &id, uint64_t iterator_
 	return iterator(id, data);
 }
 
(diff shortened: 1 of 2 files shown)


bindings/cpp/session.cpp
```diff
@@ -2465,19 +2476,29 @@ async_iterator_result session::server_send(const std::vector<dnet_raw_id> &ids,
...
 {
-	std::vector<dnet_raw_id> ids;
-	for (auto key = keys.begin(), kend = keys.end(); key != kend; ++key) {
-		dnet_id id;
-		transform(*key, id);
+	std::vector<key> keys;
+	for (auto id = ids.begin(), id_end = ids.end(); id != id_end; ++id) {
+		key k(*id);
+
+		keys.emplace_back(k);
```
abudnik: `keys.emplace_back(*id);`

bioothod: Ok, updated

PR #684, opened 2016-01-15
trans: fixed race between timeout scanning thread and completion callback

bioothod:
The latter does not remove transaction from the timer tree and transaction may be grabbed and freed by timeout checking thread.

### DIFF
--- a/library/net.c
+++ b/library/net.c
@@ -664,9 +664,16 @@ int dnet_process_recv(struct dnet_backend_io *backend, struct dnet_net_state *st
 			}
 
 			/*
-			 * Always remove transaction from 'timer' tree,
-			 * thus it will not be found by checker thread and
-			 * its callback will not be called under us
+			 * Remove transaction for the duration of callback processing,
+			 * otherwise timeout checking thread can catch up.
+			 *
+			 * Network thread also removes transaction from the tree, but network
+			 * thread can read multiple replies and put multiple packets into the IO queue,
+			 * which if processed here. Since code below inserts transaction into the timer tree
+			 * again after its callback has been completed, someone has to remove it.
+			 *
+			 * It is safe to remove transaction multiple times, but subsequent insertion will lead to crash,
+			 * if timestamp has been updated, since it is used as a key in the timer tree.
 			 */
 			dnet_trans_remove_timer_nolock(st, t);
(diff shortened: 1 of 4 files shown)


library/net.c
```diff
@@ -714,7 +713,7 @@ int dnet_process_recv(struct dnet_backend_io *backend, struct dnet_net_state *st
 			 */
 
 			pthread_mutex_lock(&st->trans_lock);
-			dnet_trans_timestamp(st, t);
+			dnet_trans_insert_timer_nolock(st, t);
```
shaitan: Now, it doesn't update 'timeout' timestamp here. It will lead to race between io thread that will handle next response for the transaction and check thread that will find and remove timeouted transactions.

What do you think about updating 'timeout' timestamp and 'timer' tree in net threads?

bioothod: This particular chunk updates timestamp after transaction has been processed. I agree, this has to be saved, since callback may take a lot of time to complete, and after transaction has been put back, it can be immediately killed with timeout, I've updated pull request.

If we update timestamp in the network thread, it will behave the same way as currently - without updating it after completion in IO thread, checking thread may grab it and free. I think that updating timestamp in network thread instead of (or better in addition to) updating it in line 666, i.e. prior calling completion callback, might be a good idea.

I will update this patch and test this idea

bioothod:
I've moved timestamp update for just received transaction into network thread, it passes all tests.

library/pool.c
```diff
@@ -254,6 +254,26 @@ static inline void make_thread_stat_id(char *buffer, int size, struct dnet_work_
...
+	if (cmd->flags & DNET_FLAGS_REPLY) {
+		struct dnet_trans *t;
+
+		/*
+		 * No need to lock @st->trans_lock, since this network thread is the only owner of the
+		 * transaction now, transaction hasn't been yet placed into IO queue.
+		 */
+		t = dnet_trans_search(st, cmd->trans);
+		if (t) {
+			dnet_trans_update_timestamp(st, t);
```
shaitan: It is not enough, it should also update 'timer' tree according to new transaction timeout.

bioothod: Yep, it should move transaction in the timer tree, I've updated pull request

PR #691, opened 2016-02-26
route: do not spam logs with debug route table content every time route table is requested

bioothod:


shaitan:
Looks good to me. Thank you.

PR #693, opened 2016-03-02
server-send: fix concealment of errors during server-send

shaitan:


example/eblob_backend.c
```diff
@@ -1078,6 +1080,10 @@ static int blob_send(struct eblob_backend_config *cfg, void *state, struct dnet_
 		if (err)
 			goto err_out_send_fail_reply;
 
+		/* stop if an error is occurred during write */
+		if (ctl->write_error)
+			goto err_out_put;
+
```
bioothod: As I understand this is needed to earlier stop of the server-send iteration over provided keys if there is write error, is that correct?

library/dnet.c
```diff
@@ -791,9 +791,6 @@ int dnet_server_send_write(struct dnet_server_send_ctl *send,
 	dnet_trans_create_send_all(s, &ctl);
 	dnet_session_destroy(s);
 
-	if (!err)
-		err = send->write_error;
-
```
bioothod: Why have you removed this error assignment?

tests/server_send.cpp
```diff
@@ -25,6 +25,8 @@
 using namespace ioremap::elliptics;
 using namespace boost::unit_test;
 
+#define ENABLE_LONG_TESTS 0
```
bioothod: It means this test will never run, should we at least provide cmake config option?
How long does it take to run this test, maybe we should turn it on always?

library/dnet.c
shaitan: Because logic around `dnet_server_send_write` expects that it returns error only if it failed to make a write and in case of error this around logic will send reply to a client with this error. This code led to duplicated reply because it made `dnet_server_send_write` to return an error when some detached write was failed, so logic around send a reply with error which repeats a reply sent by complete callback of the write.

example/eblob_backend.c
shaitan: Yep, it is. Previously, it was accomplished by https://github.com/reverbrain/elliptics/pull/693#discussion-diff-54820934, but it led to doubled reply.

tests/server_send.cpp
shaitan: It depends on internal timeout for write command used by server-send, currently this timeout is 60 seconds and can not be changed from this point.

library/dnet.c
bioothod: But if `dnet_server_send_write` doesn't return error, iterator will not stop. And `write_error` contains that kind of errors which stop iterator. And 2 errors correspond to write error and iterator error, since it will be stopped.

tests/server_send.cpp
bioothod: 60-seconds-long test doesn't look too long, I think we should enable this test

example/eblob_backend.c
bioothod: That's a very different commands.
This code changes how server-send command works, it prevents sending more keys if some of them have failed.

Code in `dnet_server_send_write` works in server-send **iterator** which stopped iterator if there are write errors. This code does not replace what you have removed.

library/dnet.c
shaitan: So, iterator's callback (`dnet_iterator_callback_server_send`) should check `write_error` too and should stop iterator if it isn't 0. I will make corresponding changes.

tests/server_send.cpp
shaitan: Ok, I will make it enabled by default and add cmake config option which will allow to disable such "long" tests.

bioothod: Sounds good to me. This option can also be used for other *long* tests when they will appear

example/eblob_backend.c
shaitan: I had missed iterator, so I have added the same check to `dnet_iterator_callback_server_send`, now it will stop iterator if some write will be failed.

tests/server_send.cpp
shaitan: I've added `DISABLE_LONG_TEST` cmake option which is disabled by default, so all long tests are enabled by default.

tests/server_send.cpp
```diff
@@ -277,6 +277,42 @@ static void ssend_test_server_send(session &s, int num, const std::string &id_pr
...
+		bool operator<(const backend &other) const {
+			if (auto cmp = dnet_addr_cmp(&addr, &other.addr))
+				return cmp < 0;
+			return backend_id < other.backend_id;
+		}
+	};
+
+	std::set<backend> backends;
+
+	for (const auto &route: s.get_routes()) {
```
bioothod: Looks like this will break ubuntu precise and other old compilers build, or is it ok?

shaitan: It builds fine on precise and, as I know, we do not support anything older than precise, do we?

bioothod: If it builds on precise, it is ok

bioothod:
Looks good to me, thank you

PR #696, opened 2016-03-11
fix for upload binary data via cpp bindings

sasokol:
I created a new constructor with data size.

### DIFF
--- a/include/elliptics/utils.hpp
+++ b/include/elliptics/utils.hpp
@@ -342,9 +342,14 @@ public:
 	}
 
 	argument_data_base(const char *data) :
-		m_data(pointer_type::from_raw(const_cast<char *>(data), std::strlen(data)))
+		m_data(pointer_type::from_raw(const_cast<char *>(data), std::strlen(data))) // It isn't correct for binary data.
 	{
 	}
+
+        argument_data_base(const char *data, size_t size) :
+                m_data(pointer_type::from_raw(const_cast<char *>(data), size))
+        {
+        }
 
 	const void *data() const
 	{


bioothod:
Looks good to me, but it requires a test, please add one which calls some method with binary data and results you read back is invalid for current code and works good with your new code.

sasokol:
Example:

```
...
char *data;
elliptics_client.write_prepare(key, "", 0,total);
uint64_t  chunk = buflen;
while ( length > 0 ) {
   ...
   if ( chunk > length ){
       chunk = length;
   } 

   FCGX_GetStr(data, chunk, request.in); // get a chunk of data from socket
   // argument_data file( data ); // In the storage bad data. But if I send to the socket a base64 encoded data then its's ok.
   argument_data file( data, chunk ); // This I use the fix.
   async_result = elliptics_client.write_plain(key, file, offset); //write the chunk into elliptick session
   offset += chunk;
....
}
```

include/elliptics/utils.hpp
```diff
@@ -342,10 +342,15 @@ class argument_data_base
 	}
 
 	argument_data_base(const char *data) :
-		m_data(pointer_type::from_raw(const_cast<char *>(data), std::strlen(data)))
+		m_data(pointer_type::from_raw(const_cast<char *>(data), std::strlen(data))) // It isn't correct for binary data.
```
shindo: > // It isn't correct for binary data.

This comment is unnecessary as the constructor is not about binary data.

include/elliptics/utils.hpp
```diff
@@ -342,10 +342,15 @@ class argument_data_base
...
 	{
 	}
 
+        argument_data_base(const char *data, size_t size) :
```
shindo: Maybe you should use here `const void *data` to explicitly specify that the constructor is for binary data.

shindo:
The pull request is not "fix" actually, as there's no bug. You just adds "helper" constructor for binary data.

Without the "fix" you can already use the constructor from `data_pointer_base`:
```
argument_data file(argument_data::pointer_type::from_raw(data, chunk));
```

sasokol:
OK, thank you. It works...
I not found the in docs and  did not understand the headers.

PR #697, opened 2016-03-15
Logs

shaitan:
I've structured logs connected with sending and handling requests in common format:
`[id]: [command string]: [action message] [transaction attributes] [other attributes]` - make it easier to view and analyze requests' logs.

Level of first start-sent and last finish-sent log message is changed to `DNET_LOG_INFO` - it allows to analyze on info level when request/response sending process starts and finishes.

I've eliminated handling lookups by cache when cache does not have requested key - avoid sending another lookup by cache directly to the backend.

### DIFF
--- a/library/trans.c
+++ b/library/trans.c
@@ -86,9 +86,9 @@ int dnet_trans_insert_nolock(struct dnet_net_state *st, struct dnet_trans *a)
 	}
 
 	if (a->st && a->st->n)
-		dnet_log(a->st->n, DNET_LOG_NOTICE, "%s: added transaction: %llu -> %s.",
-			dnet_dump_id(&a->cmd.id), (unsigned long long)a->trans,
-			dnet_addr_string(&a->st->addr));
+		dnet_log(a->st->n, DNET_LOG_NOTICE, "%s: %s: added trans: %llu -> %s/%d",
+			dnet_dump_id(&a->cmd.id), dnet_cmd_string(a->cmd.cmd), (unsigned long long)a->trans,
+			dnet_addr_string(&a->st->addr), a->cmd.backend_id);
 
 	rb_link_node(&a->trans_entry, parent, n);
 	rb_insert_color(&a->trans_entry, root);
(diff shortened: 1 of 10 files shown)


cache/slru_cache.cpp
```diff
@@ -395,13 +395,12 @@ int slru_cache_t::lookup(const unsigned char *id, dnet_net_state *st, dnet_cmd *
 	data_t* it = m_treap.find(id);
 	TIMER_STOP("lookup.find");
 
-	dnet_time timestamp;
-	memset(&timestamp, 0, sizeof(timestamp));
-
-	if (it) {
-		timestamp = it->timestamp();
+	if (!it) {
+		return -ENOTSUP;
```
bioothod: This completely breaks previous logic.
Also, this change is far out of logging changes this pull request is about.

library/trans.c
```diff
@@ -282,52 +282,36 @@ void dnet_trans_destroy(struct dnet_trans *t)
 		localtime_r((time_t *)&t->start.tv_sec, &tm);
 		strftime(str, sizeof(str), "%F %R:%S", &tm);
 
-		if ((t->command == DNET_CMD_READ) && (t->alloc_size >= sizeof(struct dnet_cmd) + sizeof(struct dnet_io_attr))) {
+		if ((t->command == DNET_CMD_READ || t->command == DNET_CMD_WRITE) &&
+		    (t->alloc_size >= sizeof(struct dnet_cmd) + sizeof(struct dnet_io_attr))) {
```
bioothod: This should not be evaluated for write command, which doesn't have `dnet_io_attr` in the reply

cache/slru_cache.cpp
shaitan: Ok, I will move this change to another PR.

shaitan: #698

library/trans.c
shaitan: `t->alloc_size` and `t` itself is not about the reply but about the request of transaction which in turn has `dnet_io_attr` for write command.

For read command there is [the code](https://github.com/reverbrain/elliptics/blob/master/library/net.c#L690) which copies `dnet_io_attr` from the reply to `t`, so when `t` is destroyed its `dnet_io_attr` will already contain info from the reply.

bioothod: Ok, makes sense

PR #698, opened 2016-03-16
Do not handle lookup by cache if it does not have the key

shaitan:
Cache handles lookups by sending another lookups directly to the backend,
but it is unnecessary when cache does not have the key.

### DIFF
--- a/cache/slru_cache.cpp
+++ b/cache/slru_cache.cpp
@@ -395,12 +395,11 @@ int slru_cache_t::lookup(const unsigned char *id, dnet_net_state *st, dnet_cmd *
 	data_t* it = m_treap.find(id);
 	TIMER_STOP("lookup.find");
 
-	dnet_time timestamp;
-	memset(&timestamp, 0, sizeof(timestamp));
-
-	if (it) {
-		timestamp = it->timestamp();
+	if (!it) {
+		return -ENOENT;
 	}
+
+	auto timestamp = it->timestamp();
 
 	guard.unlock();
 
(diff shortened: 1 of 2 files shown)


shaitan:
It shouldn't break anything.

bioothod:
Why is it unnecessary? Lookup command is about returning data from disk, it is perfectly valid not to have given key in the cache, but have it on disk.

This pull request breaks every installation which has cache enabled, since every lookup will fail if there is no data in the cache, for example after node restart.

shaitan:
I mean that "sending another lookup directly to the backend, when cache does not have the key, is unnecessary". So in this case cache should return `-ENOTSUP` to make outside logic (`dnet_process_cmd_with_backend_raw`) calls backend's handler for original request.

shaitan:
Maybe it is better to:

- return `-ENOENT` in `slru_cache_t::lookup()` what will mean that cache knows nothing about the key
- return `-ENOTSUP` in `dnet_cmd_cache_lookup` if `cache->lookup()` has returned `-ENOENT` what will mean that cache has not handled the request, so the request should be handled by something else (by backend for example).

bioothod:
Returning `-ENOTSUP` is an abuse of the error code, other levels can not determine whether cache is enabled or it doesn't have data. I think that returning `-ENOENT` and handling it properly by routing request to the backend is the way to go.

shaitan:
I've replaced `ENOTSUP` by `ENOENT` in this code and have added routing request to the backend when cache lookup is failed with `-ENOENT`.

PR #700, opened 2016-03-23
Update backend weight calculation

shaitan:
2 major changes of this PR:

- move backend weight calculation to `dnet_process_recv`. This makes time used in the calculation more correct because now it does not include time spent in user's callbacks.
- use size of whole packet instead of size of read data. Read response packet consists of `[dnet_cmd][dnet_io_attr][read data]`. First two parts have fixed size, so there is no different what to use in weight calculation: size of `read data` or size of whole packet.

### DIFF
--- a/library/node.c
+++ b/library/node.c
@@ -660,8 +660,6 @@ void dnet_update_backend_weight(struct dnet_net_state *st, const struct dnet_cmd
 		const double norm = (double)time / (double) cmd->size;
 		new_weight = 1.0 / ((1.0 / old_weight + norm) / 2.0);
 		dnet_set_backend_weight(st, cmd->backend_id, ioflags, new_weight);
-	} else {
-		new_weight = old_weight;
 	}
 	return err;
 }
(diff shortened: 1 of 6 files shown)


library/node.c
```diff
@@ -648,6 +648,24 @@ void dnet_set_backend_weight(struct dnet_net_state *st, int backend_id, uint32_t
...
+
+	const err = dnet_get_backend_weight(st, cmd->backend_id, ioflags, &old_weight);
+	if (!err &&
+	    cmd->status == 0 &&
+	    cmd->size) {
+		const double norm = (double)time / (double) cmd->size;
+		new_weight = 1.0 / ((1.0 / old_weight + norm) / 2.0);
+		dnet_set_backend_weight(st, cmd->backend_id, ioflags, new_weight);
+	} else {
+		new_weight = old_weight;
```
abudnik: Optimizing compiler will exclude this else-branch

shaitan: It was old code,  I've removed it.

PR #705, opened 2016-05-29
Dynamic backends

abudnik:
Server parses config and starts backend, which was not presented in config initially, on DNET_BACKEND_ENABLE command. Thus, it is possible to add backends on the fly.

### DIFF
--- a/library/backend.cpp
+++ b/library/backend.cpp
@@ -201,20 +201,19 @@ int dnet_backend_init(struct dnet_node *node, size_t backend_id, int *state)
 	int ids_num;
 	struct dnet_raw_id *ids;
 
-	auto &backends = node->config_data->backends->backends;
-	if (backends.size() <= backend_id) {
+	auto backend = node->config_data->backends->get_backend(backend_id);
+	if (!backend) {
 		dnet_log(node, DNET_LOG_ERROR, "backend_init: backend: %zu, invalid backend id", backend_id);
 		return -EINVAL;
 	}
 
-	dnet_backend_info &backend = backends[backend_id];
 	dnet_time start;
 	dnet_current_time(&start);
 
 	{
-		std::lock_guard<std::mutex> guard(*backend.state_mutex);
-		*state = backend.state;
-		if (backend.state != DNET_BACKEND_DISABLED) {
+		std::lock_guard<std::mutex> guard(*backend->state_mutex);
+		*state = backend->state;
+		if (backend->state != DNET_BACKEND_DISABLED) {
 			dnet_log(node, DNET_LOG_ERROR, "backend_init: backend: %zu, "
 					"trying to activate not disabled backend, elapsed: %s",
 				backend_id, elapsed(start));
@@ -230,7 +229,7 @@ int dnet_backend_init(struct dnet_node *node, size_t backend_id, int *state)
 					return -EINVAL;
 			}
 		}
-		backend.state = DNET_BACKEND_ACTIVATING;
+		backend->state = DNET_BACKEND_ACTIVATING;
 	}
 
 	dnet_log(node, DNET_LOG_INFO, "backend_init: backend: %zu, initializing", backend_id);
@@ -250,7 +249,7 @@ int dnet_backend_init(struct dnet_node *node, size_t backend_id, int *state)
 			const config backend_config = backends_config.at(index);
 			const uint32_t config_backend_id = backend_config.at<uint32_t>("backend_id");
 			if (backend_id == config_backend_id) {
-				backend.parse(&data, backend_config);
+				backend->parse(&data, backend_config);
 				found = true;
 				break;
 			}
@@ -275,39 +274,39 @@ int dnet_backend_init(struct dnet_node *node, size_t backend_id, int *state)
 		goto err_out_exit;
 	}
 
-	backend.config = backend.config_template;
-	backend.data.assign(backend.data.size(), '\0');
-	backend.config.data = backend.data.data();
-	backend.config.log = backend.log.get();
+	backend->config = backend->config_template;
+	backend->data.assign(backend->data.size(), '\0');
+	backend->config.data = backend->data.data();
+	backend->config.log = backend->log.get();
 
-	backend_io = &node->io->backends[backend_id];
+	backend_io = dnet_get_backend_io(node->io, backend_id);
 	backend_io->need_exit = 0;
-	backend_io->read_only = backend.read_only_at_start;
+	backend_io->read_only = backend->read_only_at_start;
 
-	for (auto it = backend.options.begin(); it != backend.options.end(); ++it) {
+	for (auto it = backend->options.begin(); it != backend->options.end(); ++it) {
 		const dnet_backend_config_entry &entry = *it;
-		entry.entry->callback(&backend.config, entry.entry->key, entry.value_template.data());
+		entry.entry->callback(&backend->config, entry.entry->key, entry.value_template.data());
 	}
 
-	err = backend.config.init(&backend.config);
+	err = backend->config.init(&backend->config);
 	if (err) {
 		dnet_log(node, DNET_LOG_ERROR, "backend_init: backend: %zu, failed to init backend: %d, elapsed: %s",
 			backend_id, err, elapsed(start));
 		goto err_out_exit;
 	}
 
-	backend_io->cb = &backend.config.cb;
+	backend_io->cb = &backend->config.cb;
 
-	err = dnet_backend_io_init(node, backend_io, backend.io_thread_num, backend.nonblocking_io_thread_num);
+	err = dnet_backend_io_init(node, backend_io, backend->io_thread_num, backend->nonblocking_io_thread_num);
 	if (err) {
 		dnet_log(node, DNET_LOG_ERROR, "backend_init: backend: %zu, failed to init io pool, err: %d, elapsed: %s",
 			backend_id, err, elapsed(start));
 		goto err_out_backend_cleanup;
 	}
 
-	if (backend.cache_config) {
-		backend_io->cache = backend.cache = dnet_cache_init(node, backend_io, backend.cache_config.get());
-		if (!backend.cache) {
+	if (backend->cache_config) {
+		backend_io->cache = backend->cache = dnet_cache_init(node, backend_io, backend->cache_config.get());
+		if (!backend->cache) {
 			err = -ENOMEM;
 			dnet_log(node, DNET_LOG_ERROR, "backend_init: backend: %zu, failed to init cache, err: %d, elapsed: %s",
 				backend_id, err, elapsed(start));
@@ -316,15 +315,15 @@ int dnet_backend_init(struct dnet_node *node, size_t backend_id, int *state)
 	}
 
 	ids_num = 0;
-	ids = dnet_ids_init(node, backend.history.c_str(), &ids_num, backend.config.storage_free, node->addrs, backend_id);
+	ids = dnet_ids_init(node, backend->history.c_str(), &ids_num, backend->config.storage_free, node->addrs, backend_id);
 	if (ids == NULL) {
 		err = -EINVAL;
 		dnet_log(node, DNET_LOG_ERROR, "backend_init: backend: %zu, history path: %s, "
 				"failed to initialize ids, elapsed: %s: %s [%d]",
-				backend_id, backend.history.c_str(), elapsed(start), strerror(-err), err);
+				backend_id, backend->history.c_str(), elapsed(start), strerror(-err), err);
 		goto err_out_cache_cleanup;
 	}
-	err = dnet_route_list_enable_backend(node->route, backend_id, backend.group, ids, ids_num);
+	err = dnet_route_list_enable_backend(node->route, backend_id, backend->group, ids, ids_num);
 	free(ids);
 
 	if (err) {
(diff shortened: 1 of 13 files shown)


shaitan:
```
[13:06:15][Step 1/3] Set base directory: "/tmp/buildd/elliptics-2.26.11.0/obj-x86_64-linux-gnu/tests/result/dnet_backends_test"
[13:06:15][Step 1/3] Set cocaine run directory: "/tmp/elliptics-test-run-764d4862/"
[13:06:15][Step 1/3] Starting 5 servers
[13:06:15][Step 1/3] Starting server #1
[13:06:15][Step 1/3] Started server #1
[13:06:15][Step 1/3] *** glibc detected *** /tmp/buildd/elliptics-2.26.11.0/obj-x86_64-linux-gnu/tests/../example/dnet_ioserv: corrupted double-linked list: 0x000000000096e540 ***
[13:06:15][Step 1/3] ======= Backtrace: =========
[13:06:15][Step 1/3] /lib/x86_64-linux-gnu/libc.so.6(+0x7da26)[0x2ac4bc7d6a26]
[13:06:15][Step 1/3] /lib/x86_64-linux-gnu/libc.so.6(+0x7ed0b)[0x2ac4bc7d7d0b]
[13:06:15][Step 1/3] /tmp/buildd/elliptics-2.26.11.0/obj-x86_64-linux-gnu/tests/../library/libelliptics.so.2.26(_ZNSt6vectorI25dnet_backend_config_entrySaIS0_EE13_M_insert_auxIIS0_EEEvN9__gnu_cxx17__normal_iteratorIPS0_S2_EEDpOT_+0x397)[0x2ac4bc2279d7]
[13:06:15][Step 1/3] /tmp/buildd/elliptics-2.26.11.0/obj-x86_64-linux-gnu/tests/../library/libelliptics.so.2.26(_ZN17dnet_backend_info5parseEPN7ioremap9elliptics6config11config_dataERKNS2_6configE+0xca9)[0x2ac4bc21d129]
[13:06:15][Step 1/3] /tmp/buildd/elliptics-2.26.11.0/obj-x86_64-linux-gnu/tests/../library/libelliptics.so.2.26(dnet_backend_init+0x497)[0x2ac4bc21df87]
[13:06:15][Step 1/3] /tmp/buildd/elliptics-2.26.11.0/obj-x86_64-linux-gnu/tests/../library/libelliptics.so.2.26(dnet_cmd_backend_control+0x9a8)[0x2ac4bc21fd78]
[13:06:15][Step 1/3] /tmp/buildd/elliptics-2.26.11.0/obj-x86_64-linux-gnu/tests/../library/libelliptics.so.2.26(dnet_process_cmd_raw+0x1185)[0x2ac4bc20fea5]
[13:06:15][Step 1/3] /tmp/buildd/elliptics-2.26.11.0/obj-x86_64-linux-gnu/tests/../library/libelliptics_client.so.2.26(dnet_process_recv+0x7e0)[0x2ac4bbf27de0]
[13:06:15][Step 1/3] /tmp/buildd/elliptics-2.26.11.0/obj-x86_64-linux-gnu/tests/../library/libelliptics_client.so.2.26(dnet_io_process+0x5d1)[0x2ac4bbf522c1]
[13:06:15][Step 1/3] /lib/x86_64-linux-gnu/libpthread.so.0(+0x7e9a)[0x2ac4bc543e9a]
[13:06:15][Step 1/3] /lib/x86_64-linux-gnu/libc.so.6(clone+0x6d)[0x2ac4bc84c36d]

```

tests/test_base.hpp
```diff
@@ -154,6 +159,7 @@ class config_data
 	config_data &operator() (const std::string &name, const variant &value);
 	const variant *value_impl(const std::string &name) const;
 
+	bool m_serializable;
```
shaitan: Please, add description to `m_serializable` and its getter/setter.

abudnik: done

library/backend.cpp
```diff
@@ -594,44 +631,77 @@ static int dnet_backend_set_ids(dnet_node *node, uint32_t backend_id, dnet_raw_i
...
+{
+	std::lock_guard<std::mutex> guard(backends_mutex);
+	auto it = backends.find(backend_id);
+	if (it != backends.end()) {
+		return it->second;
+	}
+	return std::shared_ptr<dnet_backend_info>();
+}
+
+void dnet_backend_info_manager::add_backend(std::shared_ptr<dnet_backend_info> &backend)
```
bioothod: Are you sure this does not leak one reference?

library/backend.h
```diff
@@ -139,14 +143,39 @@ struct dnet_backend_info
...
 {
-	std::vector<dnet_backend_info> backends;
+public:
+	/*
+	 * Locks backend with \a backend_id state mutex and fills \a status
+	 */
+	void backend_fill_status(struct dnet_node *node, struct dnet_backend_status *status, size_t backend_id) const;
+
+	std::vector<std::shared_ptr<dnet_backend_info> > get_all_backends() const
+	{
```
bioothod: Invalid indent, please put it into the same line as method declaration

library/pool.c
```diff
@@ -55,7 +55,10 @@ static void dnet_work_pool_stop(struct dnet_work_pool_place *place)
 
 	for (i = 0; i < place->pool->num; ++i) {
 		wio = &place->pool->wio_list[i];
-		pthread_join(wio->tid, NULL);
+		if (!wio->joined) {
+			pthread_join(wio->tid, NULL);
+			wio->joined = 1;
+		}
```
bioothod: This looks like a bug, how is it possible to join thread and have this structure in place so that subsequent join would sneak in?

library/pool.c
```diff
@@ -1184,46 +1197,129 @@ int dnet_io_init(struct dnet_node *n, struct dnet_config *cfg)
...
+}
+
 int dnet_server_io_init(struct dnet_node *n)
 {
 	int err;
 	size_t j = 0, k = 0;
+	size_t *backend_ids, num_backend_ids;
+	size_t backends_count;
+
+        err = dnet_get_backend_ids(n->config_data->backends, &backend_ids, &num_backend_ids);
```
bioothod: Indent problem, space instead of a tab

library/pool.c
```diff
@@ -1184,46 +1197,129 @@ int dnet_io_init(struct dnet_node *n, struct dnet_config *cfg)
 	return err;
 }
 
+int dnet_server_backend_init(struct dnet_node *n, size_t backend_id)
```
bioothod: This function looks exactly the same as code in `dnet_server_io_init()`
Also why wouldn't use tree here instead of array?

library/server.c
```diff
@@ -231,13 +231,13 @@ void dnet_server_node_destroy(struct dnet_node *n)
...
 	dnet_monitor_exit(n);
 
-	dnet_node_cleanup_common_resources(n);
-
 	dnet_route_list_destroy(n->route);
 	n->route = NULL;
 
 	dnet_backend_cleanup_all(n);
 
+	dnet_node_cleanup_common_resources(n);
```
bioothod: Looks like a bug, why is this move needed?

tests/test_base.cpp
```diff
@@ -380,6 +395,9 @@ void server_config::write(const std::string &path)
 	backends_json.SetArray();
 
 	for (auto it = backends.begin(); it != backends.end(); ++it) {
+		if (!it->is_serializable())
```
bioothod: What does this test do?

tests/test_base.cpp
```diff
@@ -562,7 +580,7 @@ std::string server_node::config_path() const
 	return m_path;
 }
 
-server_config server_node::config() const
+server_config &server_node::config()
```
bioothod: Why is this needed?

bioothod:
Please elaborate, what does this huge patch do? Why is this needed, what was the problem, why have you decided to 'fix' it this way?

library/backend.cpp
abudnik: Yep

library/pool.c
abudnik: Actually, this fix logically connected with the fix in dnet_server_node_destroy().
Previously, backends array from dnet_io struct was not freed in dnet_server_node_destroy(). Also, dnet_node_cleanup_common_resources() was called before dnet_backend_cleanup_all().
dnet_backend_cleanup_all() cleans caches and syncs caches to backends.
Now, backends array is freed in dnet_node_cleanup_common_resources(). dnet_work_pool_stop() is called from dnet_node_stop_common_resources() and subsequently from dnet_backend_cleanup_all() leading to joining same thread twice, causing crash in libpthread. This is the answer why I have added this check.
Why dnet_work_pool_stop() was not called twice before? dnet_node_cleanup_common_resources() was called before dnet_backend_cleanup_all(), so cleanups from dnet_node_cleanup_common_resources() prevented second dnet_work_pool_stop() from dnet_backend_cleanup_all().

library/server.c
abudnik: Please see long answer above.

library/pool.c
abudnik: Main reason of choosing array instead of tree - performance: backend lookup occurs every time in dnet_schedule_io().

tests/test_base.cpp
abudnik: It is needed to modify server_config. Modifying is needed to make backend serializable and update config file.

abudnik:
>why have you decided to 'fix' it this way?

backends array of dnet_backend_io structs in dnet_io was changed to array of pointers. It allows quick reallocating of the array for adding new backends on the fly, but introduces locking mechanics for every access to backends array via rw-lock.

For the same reasons backends vector in dnet_config_data was changed to a hash table encapsulated in dnet_backend_info_manager class.

library/pool.c
abudnik: fixed: https://github.com/reverbrain/elliptics/pull/705/commits/b9196974af1f9e97e819db6c10b9f0d5075a16cd

library/backend.h
abudnik: fixed: https://github.com/reverbrain/elliptics/pull/705/commits/b9196974af1f9e97e819db6c10b9f0d5075a16cd

library/pool.c
bioothod: This looks like a bug, previously pool was deleted first and never tried to be deleted again.
Pool has to be removed first before caches and backends, since pool can grab io requests from the queue and start processing them, if you've removed cache, it will crash. This is why I asked about order of `dnet_node_cleanup_common_resources()` call which you changed and which looks like a bug.

library/pool.c
bioothod: There is virtually ZERO performance issues on selecting element from tree or array of several elements. Even if you go crazy and create thousands of backends, this tiny difference is void compared to other costs.

But having a map of backends allows to add/remove them without reallocation and ugly loops with reinsertions/reallocations.

tests/test_base.cpp
bioothod: What does that mean? So far you haven't even described why is this patch needed and terminology used here is quite vague.
Please describe it in more details: why is it needed, what is this, why do you want backend 'serialization', what are you trying to achieve with these changes?

So far it looks more like a quick dirty hack to add some ugly pieces needed for obscure or no purpose.

bioothod:
There is definitely no quick reallocation, since you copy the whole array again and again when adding new elements.

Let's start with more fundamental questions: What is the purpose of this patch? What is the problem it tries to solve? Is it present in the list of issues to solve first which was introduced like a year ago?

shaitan:
This patch is the development of https://github.com/reverbrain/elliptics/pull/661.

The issue behind these changes is: allow users to add/remove backends in dynamic without restarting dnet_ioserv.

These patch introduces minimal changes to solve the issue. The next step is refactoring `dnet_io::backends` and `dnet_config_data::backends`, and merging them into one entity that will rule all backends' life. We broke solution into 2 steps because "next step" will change a lot that can be dangerous. It will be done in a separate PR.

abudnik:
>There is definitely no quick reallocation, since you copy the whole array again and again when adding new elements.

Reallocation of small array of pointers is pretty fast operation.

library/pool.c
abudnik: dnet_schedule_io() called from network thread. It is very undesirable to slowdown it, especially in case of tens of thousands RPS.

library/pool.c
abudnik: >since pool can grab io requests from the queue and start processing them, if you've removed cache, it will crash.

Pool will not grab any request, because it was stopped with dnet_node_stop_common_resources()

abudnik: Also caches & backends depends on common resources, so it seems much cleaner to me, that they are released before cleaning common resources.

tests/test_base.cpp
abudnik: From test_base.hpp:
```c++
class config_data:
...
	// it is used to prevent serialization of this object to a config (json)
	bool m_serializable;
```
server_config contains vector of backends of config_data type.
It is possible to specify backend, but it won't be serialized to a ioserv test config.
This feature used in backend_test (see test_enable_backend_after_config_change()).

library/common.cpp
```diff
@@ -1,6 +1,32 @@
...
+		return -ENOMEM;
+
+	for (size_t i = 0; i < *num_backend_ids; ++i) {
+		(*backend_ids)[i] = config_backends[i]->backend_id;
+	}
+
+	return 0;
+}
+
+struct dnet_backend_io *dnet_get_backend_io(struct dnet_io *io, size_t backend_id)
```
bioothod: This function doesn't increase reference counter or perform any other control flow operations (like remove object from array) else, why does it use locking?

library/elliptics.h
```diff
@@ -430,6 +430,7 @@ struct dnet_work_io {
 	int			thread_index;
 	uint64_t		trans;
 	pthread_t		tid;
+	int			joined;
```
bioothod: This is a sign that there is a bug somewhere which allows to call thread cleanup multiple times. It should be fixed without this flag.

library/pool.c
bioothod: You haven't described why do you need this flag? It is a bug and it should be fixed instead of introducing this binary option, which halves cleanup path to safe and unsafe parts. Please fix cleanup problem, I'm strongly against patch with such quirks.

library/pool.c
bioothod: No, this is not an excuse. Patch with this logic will not go in.
You introduce ugly hacks and bad code for strange and yet unknown reason and the only excuse to add this uglymoron is that if you add it the other way it will be a tiny bit slower? It is much simple just to drop this patch.

library/server.c
bioothod: Introducing `joined` flag and calling cleanup functions several times is a design bug, please fix this problem.

tests/test_base.cpp
bioothod: The purpose of code comments and patch description is to simplify its merge and subsequent bug fixing.
Instead of just applying the patch because it contains good description and comments you force me to think like you did when designed this code, which takes time and probably will not be 100% valid and will definitely be forgotten in a few moments, which will not help debugging other problems when the rise.

Code in `test_enable_backend_after_config_change()` gets the last backend and checks whether its initialization fails. I would never ever think that the last backend initialization has to fail. You not only introduced a patch with unknown problem and 'fix' for that unknown problem, you also hide a gem in the tests which forces the last backend fail to initialize. You didn't write a word about design of this new test, you didn't describe what does it do.
No, this is not going to work.

bioothod: Also, you break the whole logic of the test, since tests are repeatable and independent.
If it is not possible (although quite desirable), you should at least tell others what and how your tests are supposed to work and what do the test, what they expect and how this can happen in real life.

PR #708, opened 2016-07-04
json: fixed isnan() invocation in c++ code

bioothod:


bioothod:
This is needed to compile with gcc 5.3.1 which is Xenial compiler

Issue #709, opened 2016-07-05
dnet_reconnect_test failure


bioothod:
```
# Start 8 of 12: dnet_reconnect_test: 
Set base directory: "/home/zbr/awork/elliptics/build/tests/result/dnet_reconnect_test"
Set cocaine run directory: "/tmp/elliptics-test-run-10e575e3/"
Starting 1 servers
Starting server #1
Started server #1
Started servers
Running 1 test case...
/home/zbr/awork/elliptics/tests/reconnect_test.cpp(108): fatal error in "(test_failed_connection_restore, create_session(n, { 1 }, 0, 0))": (sess.request_backends_status(node.remote())), expected error: -110, received: "insufficient results count due to checker: 0 of 1 (1): No such device or address: -6"
```

[client.log.txt](https://github.com/reverbrain/elliptics/files/346920/client.log.txt)

bioothod:
Looks like #710 pull request has fixed this issue

PR #710, opened 2016-07-05
tests: fix race in reconnect test

shaitan:
Background check thread sometimes spoiled state reset triggering logic.


shaitan:
It should fix #709

PR #712, opened 2016-07-20
net: fix addicting net thread on sending responses to one state

shaitan:
If there are a lot of small responses which should be sent to some state with good connection, e.g. iterator generates such responses, net thread can get into cycle of sending these responses and will not send responses to other states until cycle is done. If responses is rather small, the cycle will be done only when all responses are sent.

I've added break to this cycle if a response is fully sent, so net thread will switch to another state for sending response to it.

### DIFF
--- a/library/pool.c
+++ b/library/pool.c
@@ -689,6 +689,7 @@ static int dnet_process_send_single(struct dnet_net_state *st)
 
 			dnet_io_req_free(r);
 			st->send_offset = 0;
+			break;
 		}
 
 		if (err)


bioothod:
Should this problem be handled by having more network threads?

This patch forces elliptics to go to kernel, to reschedule process and so on - just to send a single packet. And if there are multiple packets in the queue, this will require multiple epoll schedule roundtrip.

shaitan:
>Should this problem be handled by having more network threads?

only if each state including accepter will have separate thread otherwise iteration spams one net thread, so it stops serving other states.

shaitan:
We could add some limit on number of responses sent in a row to one state, so instead of breaking after each response it will break only if quite a lot responses are sent to some state But I think we should proceed from reality and measure how it will affect to iteration because other commands doesn't produce such number of small responses.

